PR #3, opened 2018-05-24
Add satellite streaming API.

anuraaga:
Some style for documentation, etc follows the pattern in this file

https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub.proto

The major difference vs the PRD is that I left the frame header as opaque bytes - modeling it in proto would be difficult since proto only supports 32-bit fields, and I think it's used rarely enough (not used by Fleet) that we can leave it like that.

Also, when discussing with Ricardo, he suggested `framing` is easier to understand than `encoding` for users. This is good since it's easier to understand for me too.

Please feel free to leave comments on fine details of the documentation as well as overall API design.


api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -0,0 +1,131 @@
...
+  AX25 = 1;
+}
+
+// A response from the `OpenSatelliteStream` method.
+message SatelliteStreamResponse {
+  // A response on the strema. Currently the only type of response is telemetry received from the
+  // satellite.
+  oneof Response {
+    // A response from a satellite containing telemetry.
+    ReceiveTelemetryResponse receive_telemetry_response = 1;
```
gavinreaney: Assuming I open a stream at time T, will this response then include only telemetry received after time T?

As a client of the API, am I expected to continually read the telemetry responses? What happens if the client is not calling receive very often? Does the telemetry end up being buffered in the server and how much buffering do we allow? I guess we should say something in the docs about what we expect from clients that wish to receive telemetry and the impact of slow/infrequent client reads.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -0,0 +1,131 @@
...
+  }
+}
+
+// A response from a satellite containing telemetry.
+message ReceiveTelemetryResponse {
+  // The ID of the plan on which the telemetry was received.
+  string plan_id = 1;
+
+  // The telemetry received.
+  repeated Telemetry telemetry = 2;
```
gavinreaney: Does this need to be repeated? I guess telemetry can still be split across SatelliteStreamResponse messages anyway.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -0,0 +1,131 @@
...
+import "google/protobuf/timestamp.proto";
+
+package stellarstation.api.v1;
+
+option java_multiple_files = true;
+option java_outer_classname = "StellarstationProto";
+option java_package = "com.stellarstation.api.v1";
+
+// The public API service of Stellarstation, a system supporting communication between satellites
+// and antennas. This API is for use by an operator of a satellite - it is illegal to specify
```
kenichi-fukushima: I recommend you don't mention anything related to laws and regulations unless you are 100% sure. Also it is not the client but the server who should ensure compliance to laws and regulations.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -0,0 +1,131 @@
...
+  // raw IQ data.
+  RAW = 0;
+
+  // AX.25 (Amateur X.25) framing - https://en.wikipedia.org/wiki/AX.25
+  AX25 = 1;
+}
+
+// A response from the `OpenSatelliteStream` method.
+message SatelliteStreamResponse {
+  // A response on the strema. Currently the only type of response is telemetry received from the
```
kenichi-fukushima: typo: strema -> stream

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -0,0 +1,131 @@
...
+  // Open a stream to a satellite. The returned stream is bi-directional - it can be used by the
+  // client to send commands to the satellite and data received from the satellite will be returned
+  // as it is made available. All telemetry received from the satellite will be returned as soon as
+  // this method is called.
+  // If the satellite is not found or the client is not authorized for it, the stream will be closed
+  // with a `NOT_FOUND` error.
+  rpc OpenSatelliteStream (stream SatelliteStreamRequest) returns (stream SatelliteStreamResponse);
+}
+
+// Request for the `OpenSatelliteStreaam` method.
```
kenichi-fukushima: Typo: Streaam -> Stream

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
anuraaga: Yup you're right I should have just mentioned passes

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
anuraaga: Yeah I was debating on this - I was thinking of maintaining consistency with `SendCommandRequest`, which does need the ability to send multiple frames, but here I guess singular makes more sense.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -0,0 +1,131 @@
...
+  // No framing. For sent payloads, this is usually pre-framed data while for received ones, it is
+  // raw IQ data.
+  RAW = 0;
+
+  // AX.25 (Amateur X.25) framing - https://en.wikipedia.org/wiki/AX.25
+  AX25 = 1;
+}
+
+// A response from the `OpenSatelliteStream` method.
+message SatelliteStreamResponse {
```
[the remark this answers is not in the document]
anuraaga: We decided that responses to commands are implemented by the satellite / protocol, not the API, so users would just look at received telemetry, and we don't need to return a response for a command. I agree it's weird, but not sure if there's a clear way to document that here either since it's about things *not* in the API. What do you think?

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
anuraaga: Oops - I had forgotten to add `stream_id` and semantics of resumption. Let me know if the new documentation is still unclear.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -0,0 +1,154 @@
...
+import "google/protobuf/timestamp.proto";
+
+package stellarstation.api.v1;
+
+option java_multiple_files = true;
+option java_outer_classname = "StellarstationProto";
+option java_package = "com.stellarstation.api.v1";
+
+// The public API service of Stellarstation, a system supporting communication between satellites
+// and antennas. This API is for use by an operator of a satellite - it is illegal to specify
```
hoshir: A word "Ground Station" would be better than "antenna" since antenna means physical object while ground station means entire system that has capability to receive radio from satellites. Ricardo recommends the word "ground station" as well.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -0,0 +1,154 @@
...
+
+// A response from the `OpenSatelliteStream` method.
+message SatelliteStreamResponse {
+  // The ID of this stream, which can be used in a subsequent call to `OpenSatelliteStream` to
+  // resume this stream on interruption (e.g., if a network failure causes the connection to be
+  // closed mid-stream). Telemetry received on a previously opened stream with no current listener
+  // will be buffered for a maximum of 1 minute, and the stream itself will be cancelled and no more
+  // buffering will occur if no client resumes the stream with `OpenSatelliteStream` within 10
+  // minutes.
+  string stream_id = 1;
```
hoshir: Could you add a comment if this id is unique among all customers or not since that information is important for Mission Control Software vendor who will handle multiple customers.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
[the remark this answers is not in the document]
gavinreaney: OK, understood.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
gavinreaney: The comment indicates that as long as there is a listener, then we'll buffer the data. Presumably there is still some limit in the case that the client doesn't read the stream fast enough?

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -0,0 +1,154 @@
...
+  // and returned. When the client is done with the stream, it should close it successfully. If the
+  // stream is cancelled with an unexpected error at any point, telemetry received before the next
+  // call to `OpenSatelliteStream` to resume the stream will be buffered up to 1 minute, and if no
+  // call to `OpenSatelliteStream` is made with `stream_id` specified to resume the stream, it will
+  // expire in 10 minutes and a call to `OpenSatelliteStream` with that `stream_id` will be closed
+  // with a `ABORTED` error.
+  //
+  // If the satellite is not found or the client is not authorized for it, the stream will be closed
+  // with a `NOT_FOUND` error.
+  rpc OpenSatelliteStream (stream SatelliteStreamRequest) returns (stream SatelliteStreamResponse);
```
gavinreaney: Dumb question :) What happens if I open a stream to a satellite but there is no active plan at the moment? I assume this whole API is based around the idea that the client only calls it when a plan has started?

gavinreaney: Another dumb question: what happens if I (or another user) tries to open a stream to the same satellite? Is that allowed?

anuraaga: Nah the API is just to get any data for a satellite as it arrives. It's also why I mentioned the stream can be left open for any amount of time - it would be idle until a plan started but it's ok. Tried adding some words to see if it's clearer that way.

anuraaga: Yeah it's allowed - not sure if it's worth mentioning that explicitly since it's lack of a restriction rather than a restriction.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
anuraaga: All IDs are globally unique - I'd rather not mention that for every type of ID we have. Do you think we need to?

anuraaga: It says there will be buffering when there is `no current listener` ;) Data is returned on the stream right away without buffering.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -0,0 +1,150 @@
...
+import "google/protobuf/timestamp.proto";
+
+package stellarstation.api.v1;
+
+option java_multiple_files = true;
+option java_outer_classname = "StellarstationProto";
+option java_package = "com.stellarstation.api.v1";
+
+// The public API service of Stellarstation, a system supporting communication between satellites
+// and groundstations. This API is for use by an operator of a satellite - it is illegal to specify
```
kenichi-fukushima: Remove the phrase `it is illegal...`? Maybe you use the term "illegal" not as violation of laws but as invalid parameters?

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -0,0 +1,150 @@
...
+  oneof Response {
+    // A response from a satellite containing telemetry.
+    ReceiveTelemetryResponse receive_telemetry_response = 2;
+  }
+}
+
+// A response from a satellite containing telemetry.
+message ReceiveTelemetryResponse {
+  // The ID of the plan on which the telemetry was received.
+  string plan_id = 1;
```
kenichi-fukushima: As per the offline discussion, we decided we don't support use cases where the user reserves overlapping passes. Is this field needed then?

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -0,0 +1,150 @@
...
+
+// The public API service of Stellarstation, a system supporting communication between satellites
+// and groundstations. This API is for use by an operator of a satellite - it is illegal to specify
+// satellites that the operator does not own or passes they have not reserved.
+// Only using IDs shown on the StellarStation Console or returned in API responses will ensure all
+// inputs are valid.
+service StellarStationService {
+  // Open a stream to a satellite. The returned stream is bi-directional - it can be used by the
+  // client to send commands to the satellite and data received from the satellite will be returned
+  // as it is made available. All telemetry received from the satellite on reserved plans from this
```
kenichi-fukushima: `reserved plans` -> `reserved passes`? I'm arguing we will end up having the concept of plan in public APIs but we haven't decided.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -0,0 +1,150 @@
...
+// A request to send commands to a satellite.
+message SendSatelliteCommandsRequest {
+  // The framing to be applied to `command` before transmission. If `command` has already been
+  // framed, this should be set to RAW and no additional processing will take place before
+  // modulation. When applying framing, each `command` will be treated as a single frame. Framing
+  // parameters used will be the ones registered for the satellite in the StellarStation Console.
+  Framing output_framing = 1;
+
+  // The command frames to send to the satellite. All commands will be transmitted in sequence
+  // immediately, during which time telemetry will not be received for the given `plan_id`. After
```
kenichi-fukushima: Remove mention of plan_id.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
anuraaga: Indeed I was thinking `IllegalArgumentException`, but since this gRPC it was better to think of `INVALID_ARGUMENT` :)

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
anuraaga: Yup you're right we shouldn't mention it anywhere yet until we support multiple simultaneous passes.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// A type of framing of a binary payload used in satellite communicaation.
+enum Framing {
+  // No framing. For sent payloads, this is usually pre-framed data while for received ones, it is
+  // raw IQ data.
+  RAW = 0;
+
+  // AX.25 (Amateur X.25) framing - https://en.wikipedia.org/wiki/AX.25
+  AX25 = 1;
```
reiinakano: If we specify AX25 during transmission, don't we need some mechanism to allow the user to set the desired AX25 address field, control field, and PID field? We could have reasonable defaults if they don't care, but really, the only fields that aren't user-specific here are the start/end flags and checksum fields.

anuraaga: Yeah the parameters are registered into the DB, rather than being in the API, documented here

https://github.com/infostellarinc/stellarstation-api/blob/master/api/src/main/proto/stellarstation/api/v1/stellarstation.proto#L78

Ricardo thinks this should be fine for conceivable use cases, if something comes up to require it in the API we can add it later. Does that make sense?

reiinakano: Oh, there it is. :) Makes sense.

PR #7, opened 2018-05-30
Add a fake server and client to allow easy experimenting with API stu…

anuraaga:
…b and demonstrate simple code usage.

I'll later package up a docker image that can run fakeserver, as well as extract the API key for writing test client code.


examples/fakeserver/src/main/java/com/stellarstation/api/fakeserver/FakeStellarStationService.java
```diff
@@ -0,0 +1,125 @@
...
+import java.time.Duration;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
+import javax.inject.Inject;
+
+/**
+ * A fake implementation of {@link StellarStationServiceImplBase} which:
+ *
+ * <ul>
+ *   <li>Returns 1MB of random telemetry payload every second.
```
kenichi-fukushima: Isn't this comment for `openSatelliteStream`?

examples/fakeserver/src/main/java/com/stellarstation/api/fakeserver/FakeStellarStationService.java
```diff
@@ -0,0 +1,125 @@
...
+  @Inject
+  FakeStellarStationService() {}
+
+  @Override
+  public StreamObserver<SatelliteStreamRequest> openSatelliteStream(
+      StreamObserver<SatelliteStreamResponse> responseObserver) {
+    ServiceRequestContext ctx = RequestContext.current();
+    ctx.setRequestTimeout(Duration.ZERO);
+    ctx.setMaxRequestLength(0);
+    var future =
```
kenichi-fukushima: Can you use an explicit type unless that's too verbose?

examples/printing-client/src/main/java/com/stellarstation/api/examples/printingclient/PrintingClientMain.java
```diff
@@ -0,0 +1,117 @@
...
+    requestStream.onNext(SatelliteStreamRequest.newBuilder().setSatelliteId("5").build());
+
+    ScheduledExecutorService commandExecutor = Executors.newScheduledThreadPool(1);
+
+    // Send some arbitrary commands every 3 seconds.
+    commandExecutor.scheduleAtFixedRate(
+        () -> {
+          requestStream.onNext(
+              SatelliteStreamRequest.newBuilder()
+                  .setSatelliteId("5")
```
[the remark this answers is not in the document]
anuraaga: Yeah - I was thinking it's a good sanity check to leave it in every request since it's not really much overhead, and it reduces a "branch" within the API. What do you think?

examples/fakeserver/src/main/java/com/stellarstation/api/fakeserver/FakeStellarStationService.java
anuraaga: I wanted to give an idea of this fake server all in once place, and in practice there is only one method for now. If this fake server gets more complicated may need to change it but I think it's appropriate here for now.

examples/fakeserver/src/main/java/com/stellarstation/api/fakeserver/FakeStellarStationService.java
anuraaga: I wasn't sure what the type is other than "something with a cancel method", so thought it's a good use case of `var`. But if it's important for readers than makes sense to be explicit (I guess very few people know what a `ScheduledFuture` is though).

examples/fakeserver/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Verifies a fake API key
+- Returns 1MB of random telemetry every second
+- Echos back received commands
+- Cancels the connection after 5 minutes
+- Only allows requests for satellite ID `"5"`, other IDs show behavior for non-existent satellites
+
+## Running
+
+// TODO(rag): Add instructions for running with docker after pushing an image.
```
hoshir: Examples should be as simple as possible so that users can focus on learning the APIs which is new to them. As a command to run the server, just using gradle task or Java command would better not to distract users by using additional technical stack like Docker, which could confuse them on what is doing or may need additional steps of installing docker.

For example, we could add execute task in gradle and give an instruction like this.

```
task execute(type:JavaExec) {
    main = 'com.stellarstation.api.fakeserver.FakeServerMain'
    classpath = sourceSets.main.runtimeClasspath
}
```

> Run following command at the top directory of the repository in order to launch the fake server.
> $ ./gradlew examples:fakeserver:execute

anuraaga: Thanks - there is already a `run` task which indeed I should document here. I'll also document docker since the intention is people can run the fakeserver without cloning this git repository during their initial development phase.

PR #11, opened 2018-06-06
Add links to other language ways to authenticate.

anuraaga:
/cc @kenichi-fukushima adding a link to pointers on authenticating a C++ client :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,13 @@ StellarStationServiceStub client =
         .withCallCredentials(MoreCallCredentials.from(credentials));
 ```
 
+Other languages have similar methods for loading Service Account JWT Access Credentials.
+For example,
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -66,6 +66,13 @@ StellarStationServiceStub client =
         .withCallCredentials(MoreCallCredentials.from(credentials));
 ```
 
+Other languages have similar methods for loading Service Account JWT Access Credentials.
+For example,
+
+C++ - https://github.com/grpc/grpc/blob/583f39ad94c0a14a50916e86a5ccd8c3c77ae2c6/include/grpcpp/security/credentials.h#L144
```
hoshir: Shouldn't it be a link for the code in master branch?

anuraaga: Oops actually I meant to use permalinks for them all - I want to make sure `master` changes don't end up pointing to the wrong place or it'd be too confusing, fixed. Hopefully we'll replace the links with small code snippets going forward :)

PR #12, opened 2018-06-11
Fix definition of RAW framing.

anuraaga:
My understanding of what it means to be raw or not was wrong I think. I guess it is actually IQ -(demodulation)-> RAW -(deframe)-> AX.25 Frames.

@reiinakano does this make sense?


reiinakano:
For transmission, this API is fine I think, although maybe RAW should be named BITSTREAM? 

For reception, I don't know if there's much value in providing BITSTREAM after demodulation. I don't know our customers, but I think it makes more sense to provide the raw IQ or the already-decoded frames. Also, if we decide to provide/store BITSTREAM after demodulation, we'd need a lot more storage space and network speed per pass.

PR #19, opened 2018-06-14
Build and publish python stubs

anuraaga:


build.gradle
```diff
@@ -107,3 +107,18 @@ task wrapper(type: Wrapper) {
     gradleVersion = '4.8'
     distributionType = 'ALL'
 }
+
+ext.condaRoot = gradle.gradleUserHomeDir.toPath().resolve('curiostack/python/bootstrap/miniconda2-4.5.4')
```
hoshir: Miniconda or Conda is a distribution for data science. While it is a quite famous for the purpose, it is not intended for general pupose. I would recommend to use standard python and virtualenv to create isolated environment.

anuraaga: Unfortunately it is annoying to install python in a cross-platform way - using conda allows creating repos that will build in any environment which is why I picked using it. This is only used by Gradle when building, it's not used by users so I think it's fine.

hoshir: Understood. I hope we will have easier way of installing python in cross platform than conda.

PR #20, opened 2018-07-12
Add IQ as a type of payload.

anuraaga:


reiinakano:
To use this to transmit, we'll need to know its sampling rate and format(float, int8, etc). Also, IQ data containing one packet to be sent is likely to be bigger than 1MB, so additional chunking might be necessary.

anuraaga:
Right, it seemed complicated so for now I clarified in the comments that `IQ` is read only

PR #24, opened 2018-07-26
Fix TLE comment.

anuraaga:
Since we will return fresher TLE for better tracking where possible, it is not necessarily the same TLE that was used to compute the plan.


anuraaga:
By the way, separated out message for TLE.

PR #38, opened 2018-08-16
Publish C++ stubs.

anuraaga:
I will publish to `stable` and then update README after @reiinakano gets a chance to try these out.

### DIFF
--- a/stubs/cpp/package/conanfile.py
+++ b/stubs/cpp/package/conanfile.py
@@ -1,7 +1,7 @@
 from conans import ConanFile, CMake
 
 
-class StellarstationapiConan(ConanFile):
+class StellarstationApiConan(ConanFile):
     name = "stellarstation-api"
     version = "0.0.8"
     license = "Apache-2.0"
--- a/stubs/cpp/package/conanfile.py
+++ b/stubs/cpp/package/conanfile.py
@@ -1,0 +1,35 @@
+from conans import ConanFile, CMake
+
+
+class StellarstationapiConan(ConanFile):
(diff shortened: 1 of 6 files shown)


stubs/cpp/package/conanfile.py
```diff
@@ -0,0 +1,35 @@
+from conans import ConanFile, CMake
+
+
+class StellarstationapiConan(ConanFile):
```
hoshir: Cases of class name looks not usual. Is this intentional?

stubs/cpp/package/conanfile.py
```diff
@@ -0,0 +1,35 @@
...
+    def package(self):
+        self.copy("*.h", dst="include", src="src")
+        self.copy("*.lib", dst="lib", keep_path=False)
+        self.copy("*.dll", dst="bin", keep_path=False)
+        self.copy("*.dylib*", dst="lib", keep_path=False)
+        self.copy("*.so", dst="lib", keep_path=False)
+        self.copy("*.a", dst="lib", keep_path=False)
+
+    def package_info(self):
+        self.cpp_info.libs = ["hello"]
```
hoshir: Is this correct?

stubs/cpp/package/conanfile.py
anuraaga: It was automatically generated by conan, but seems like we can rename it.

stubs/cpp/package/conanfile.py
anuraaga: Thanks!

PR #39, opened 2018-08-17
Separate GroundStationService from StellarStationService

anuraaga:
The API will be clearer for users if we separate satellite operator APIs from ground station APIs since the users are usually separate. It also allows us to tailor the concepts to each use case as needed (see my Plan change in https://github.com/infostellarinc/stellarstation-api/commit/9c32665ad52c929a306f1f305f9b295229b6aec8).

Luckily, there was already a comment on `StellarStationService` saying it's for satellite operators, putting `ListGroundStationPlans` there wasn't appropriate.

Also removed a file that prevents running fakeserver in IntelliJ.


api/src/main/proto/stellarstation/api/v1/groundstation.proto
```diff
@@ -0,0 +1,107 @@
...
+
+// The public API of StellarStation for use by ground station operators integrating with the
+// StellarStation ground network. It is invalid to specify ground station IDs that the operator
+// does not own. Only using IDs shown on the StellarStation Console or returned in API responses
+// will ensure all inputs are valid.
+//
+// A plan is a scheduled pass that will be executed to send and receive data between the ground
+// station and satellite during the time range.
+//
+// A plan is a scheduled pass that will be executed to send and receive data between the ground
```
hoshir: Duplicated.

api/src/main/proto/stellarstation/api/v1/groundstation.proto
```diff
@@ -0,0 +1,107 @@
...
+// station and satellite during the time range.
+service GroundStationService {
+  // Lists the plans for a particular ground station.
+  //
+  // The request will be closed with an `INVALID_ARGUMENT` status if `ground_station_id`,
+  // `aos_after`, or `aos_before` are missing, or the duration between the two times is longer than
+  // 31 days.
+  rpc ListGroundStationPlans (ListGroundStationPlansRequest) returns (ListGroundStationPlansResponse);
```
hoshir: Since this proto is for ground stations, what do you think of naming it as ListPlans?

api/src/main/proto/stellarstation/api/v1/groundstation.proto
anuraaga: Oops fixed

api/src/main/proto/stellarstation/api/v1/groundstation.proto
anuraaga: Good idea, done.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -44,6 +44,8 @@ service StellarStationService {
   // The request will be closed with an `INVALID_ARGUMENT` status if `ground_station_id`,
   // `aos_after`, or `aos_before` are missing, or the duration between the two times is longer than
   // 31 days.
+  //
+  // Deprecated - use GroundStationService.ListGroundStationPlans
```
hoshir: Need to fix as well: GroundStationService.ListPlans

anuraaga: Fixed

PR #43, opened 2018-08-17
Add ListUpcomingAvailablePasses API definition.

anuraaga:


api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -40,6 +40,15 @@ option java_package = "com.stellarstation.api.v1";
 // station and satellite during the time range.
 service StellarStationService {
 
+  // Lists upcoming passes for a satellite. Passes that can be scheduled for the satellite with an
+  // Acquisition of Signal (AOS) within 14 days from now will be returned. This method does not
+  // reserve passes in any way - it is possible for a pass that is returned by this method to be
+  // unschedulable when actually trying to schedule because a conflict appeared during that time.
+  //
+  // If the satellite is not found or the client is not authorized for it, the stream will be closed
```
kenichi-fukushima: This API is not streaming.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -166,3 +175,42 @@ message Telemetry {
...
+// Response for the `ListUpcomingAvailablePasses` method.
+message ListUpcomingAvailablePassesResponse {
+  // The upcoming, schedulable passes for the satellite.
+  repeated Pass pass = 1;
+}
+
+// A pass during which a satellite can be communicated with from a given ground station.
+message Pass {
+  // A unique ID for this pass. This can be used in other APIs that accept `pass_id`.
+  string id = 1;
```
kenichi-fukushima: I'd like to avoid giving users illusion that passes are identifiable objects. Even passes that are off by 1 seconds from each other have different ids, which may surprise users. Provided that `ReservePass` will be the only API that takes this value in the short term, I would make this explicitly a pass reservation token.

PR #47, opened 2018-08-21
Improvements to python stub upload

gavinreaney:
1) Grab the version number directly from the API project (this required making the setup.py a simple template).

2) Only do the upload if the version is not a snapshot

3) Skip the version if it already exists rather than failing

4) Take pypi username and password from gradle properties


stubs/python/build.gradle
```diff
@@ -58,17 +58,31 @@ task fillRunProtocScript {
     }
 }
 
+def apiVersion = project.project(":api").property("version").toString()
+
+task prepareSetupPy {
```
anuraaga:
```groovy
inputs.property(apiVersion);
inputs.file('....../setup.py.tmpl')
outputs.file('...../setup.py')
```

stubs/python/build.gradle
```diff
@@ -97,7 +114,9 @@ afterEvaluate {
         dependsOn buildPythonPackage
 
         executable 'bash'
-        args '-c', condaCommand('twine upload dist/*')
+        args '-c', condaCommand("twine upload -u $pypi_user -p $pypi_password dist/*")
```
anuraaga: I think properties need to come from `rootProject` to be overridable from command line (we may want to do that for whatever reason).

Also let's check the properties

```groovy
def user = rootProject.findProperty('pypiUser')
if (!user) {
  throw
}
```

etc.

stubs/python/build.gradle
```diff
@@ -58,17 +58,31 @@ task fillRunProtocScript {
     }
 }
 
+def apiVersion = project.project(":api").property("version").toString()
```
anuraaga: I'm not sure but I think if we put the version in the top level gradle.properties, it will be used in subprojects. Can you check if that's true?

stubs/python/build.gradle
gavinreaney: For this case I realized we can't throw an exception here because it runs during the config phase.

I'll try making this a generic task that calls exec instead.

gradle.properties
```diff
@@ -15,3 +15,4 @@
 #
 
 org.curioswitch.curiostack.version = 0.0.143
+apiVersion = 0.0.9-SNAPSHOT
```
gavinreaney: Note: I renamed 'version' to 'apiVersion'. I didn't see anything using the existing property though.

anuraaga: Does this need to be `apiVersion` and not `version`? I don't think the java will publish correctly without a property called `version`

anuraaga: Ah - it was implicitly used by the java publishing :)

gavinreaney:
Verified that version number is reflected in the outputs used for java publishing.

PR #48, opened 2018-08-22
Replace frequency in plans with detailed protocol information which i…

anuraaga:
…s required for actually configuring ground stations.

Frequency is generally not enough to configure a ground station for communication with a satellite, so return all the configuration instead.

### DIFF
--- a/api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
+++ b/api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
@@ -92,11 +92,11 @@ message Plan {
 
   // Configuration of the radio device used for downlinking from the satellite. Ground stations will
   // need to configure reception during the plan to match this device.
-  radio.RadioDeviceConfiguration satellite_downlink_radio_device = 5;
+  radio.RadioDeviceConfiguration downlink_radio_device = 5;
 
   // Configuration of the radio device used for uplinking to the satellite. Ground stations will
   // need to configure transmission during the plan to match this device.
-  radio.RadioDeviceConfiguration satellite_uplink_radio_device = 6;
+  radio.RadioDeviceConfiguration uplink_radio_device = 6;
 }
 
 // Unparsed TLE data for a satellite - https://en.wikipedia.org/wiki/Two-line_element_set
--- a/api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
+++ b/api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
@@ -90,13 +90,13 @@ message Plan {
   // The time of LOS between the ground station and satellite in this plan.
   google.protobuf.Timestamp los_time = 4;
 
-  // Specification of the radio device used for downlinking from the satellite. Ground stations will
+  // Configuration of the radio device used for downlinking from the satellite. Ground stations will
   // need to configure reception during the plan to match this device.
-  radio.RadioDevice satellite_downlink_radio_device = 5;
+  radio.RadioDeviceConfiguration satellite_downlink_radio_device = 5;
 
-  // Specification of the radio device used for uplinking to the satellite. Ground stations will
+  // Configuration of the radio device used for uplinking to the satellite. Ground stations will
   // need to configure transmission during the plan to match this device.
-  radio.RadioDevice satellite_uplink_radio_device = 6;
+  radio.RadioDeviceConfiguration satellite_uplink_radio_device = 6;
 }
 
 // Unparsed TLE data for a satellite - https://en.wikipedia.org/wiki/Two-line_element_set
--- a/api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
+++ b/api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
@@ -17,6 +17,7 @@
 syntax = "proto3";
 
 import "google/protobuf/timestamp.proto";
+import "stellarstation/api/v1/radio/radio.proto";
 
 package stellarstation.api.v1.groundstation;
 
@@ -89,13 +90,13 @@ message Plan {
   // The time of LOS between the ground station and satellite in this plan.
   google.protobuf.Timestamp los_time = 4;
 
-  // The center frequency, in Hz, for downlinking in this plan. 0 if downlink is not available in
-  // this plan.
-  uint64 downlink_center_frequency_hz = 5;
+  // Specification of the radio device used for downlinking from the satellite. Ground stations will
+  // need to configure reception during the plan to match this device.
+  radio.RadioDevice satellite_downlink_radio_device = 5;
 
-  // The center frequency, in Hz, for uplinking in this plan. 0 if uplink is not available in this
-  // plan.
-  uint64 uplink_center_frequency_hz = 6;
+  // Specification of the radio device used for uplinking to the satellite. Ground stations will
+  // need to configure transmission during the plan to match this device.
+  radio.RadioDevice satelltie_uplink_radio_device = 6;
 }
 
 // Unparsed TLE data for a satellite - https://en.wikipedia.org/wiki/Two-line_element_set
--- a/api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
+++ b/api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
@@ -96,7 +96,7 @@ message Plan {
 
   // Specification of the radio device used for uplinking to the satellite. Ground stations will
   // need to configure transmission during the plan to match this device.
-  radio.RadioDevice satelltie_uplink_radio_device = 6;
+  radio.RadioDevice satellite_uplink_radio_device = 6;
 }
 
 // Unparsed TLE data for a satellite - https://en.wikipedia.org/wiki/Two-line_element_set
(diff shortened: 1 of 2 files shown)


api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -0,0 +1,92 @@
...
+
+// Messages for modeling radio devices (e.g., receivers, transmitters) in API messages.
+
+// A radio device for RF communication. Usually, a satellite will have one or more radio devices for
+// communicating with the ground and a ground station will have one or more radio devices for
+// communicating with the satellite. Such devices may not actually correspond to actual hardware
+// devices, i.e., in the case of Software Defined Radio (SDR).
+message RadioDevice {
+  // The center frequency of the device, in Hz.
+  uint64 center_frequency_hz = 1;
```
reiinakano: If we have `RadioDevice` represent a ground station's radio device, then I'm not sure a single `center_frequency_hz` is appropriate. Ground stations will probably have a range of frequencies they support, unlike satellites which have a few discrete frequencies.

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -0,0 +1,92 @@
...
+
+  // The protocol used by this device when doing RF communication.
+  CommunicationProtocol protocol = 2;
+}
+
+// A communication protocol used with a radio device. These must contain all the parameters
+// needed to configure the device for use either in transmission or reception.
+message CommunicationProtocol {
```
reiinakano: Are we going to handle the special cases like APT and CW here (that define both modulation and encoding in their standard)? 

Also, for higher bands, I think it's common to use a combination of error correction algorithms like Viterbi and Reed Solomon, not just one.

I guess we could handle those moving forward based on need.

anuraaga: Yeah I think we can add them later as needed, I think it'll just be comment tweaking (e.g., modulation not filled for APT).

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
reiinakano: Hmm, my impression was that the Lithum radio, for example, corresponds to actual hardware with a range of frequencies it supports. I guess my question would be in those cases, what exactly does the single `center_frequency_hz` represent?

kenichi-fukushima: Maybe we would rename the message `RadioDeviceConfiguration`

anuraaga: Thanks, that actually clears it up great I think

kenichi-fukushima:
LGTM. I would defer final approval to Rei, Ricardo and Ryuichi.

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
reiinakano: Ah, now I get it :)

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -89,13 +90,13 @@ message Plan {
...
+  // Configuration of the radio device used for downlinking from the satellite. Ground stations will
+  // need to configure reception during the plan to match this device.
+  radio.RadioDeviceConfiguration satellite_downlink_radio_device = 5;
 
-  // The center frequency, in Hz, for uplinking in this plan. 0 if uplink is not available in this
-  // plan.
-  uint64 uplink_center_frequency_hz = 6;
+  // Configuration of the radio device used for uplinking to the satellite. Ground stations will
+  // need to configure transmission during the plan to match this device.
+  radio.RadioDeviceConfiguration satellite_uplink_radio_device = 6;
```
hoshir: Should this be ground_station_uplink_radio_device?
I think just `uplink_radio_device` and `downlink_radio_device` works without `satellite_` prefix.

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -0,0 +1,92 @@
...
+    Ax25 ax25 = 3;
+  }
+}
+
+// Communication of a stream of bits with no framing defined.
+message BitStream {
+}
+
+// PACKET communication based on AX.25. https://www.sigidwiki.com/wiki/PACKET.
+message Ax25 {
```
hoshir: AX25 seems better if it matches proto naming rule.

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// Communication of a stream of bits with no framing defined.
+message BitStream {
+}
+
+// PACKET communication based on AX.25. https://www.sigidwiki.com/wiki/PACKET.
+message AX25 {
+  // Bit rate, e.g. 9600 for FSK9600, 1200 for AFSK1200.
+  uint64 bit_rate = 1;
```
rtubio: Two things regarding this field in the AX25 data type:

1, major) bitrate is a parameter of the modulator that should be part of the radio configuration object. AX.25 is just an encoding format that runs on top of several modulations with different parameters, including different bitrates. Please use the following [https://www.tapr.org/pdf/AX25.2.2.pdf] as a reference.

2, minor) it is more common to see "bitrate" alltogether in terms of communications naming conventions.

anuraaga: Thanks - can you take another look to see if the fieldname / comment make sense or could be more clear? @rtubio @reiinakano

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
rtubio: In this structure, Modulation is a parameter for the radio configuration - we should move it from here to the radio configuration device.

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
rtubio: The rest of the message LGTM.

rtubio: The comment makes sense, although I would refer to the official specification for AX.25 better. We just need to remove the reference to "bit_rate" within the AX.25 protocol header.

anuraaga: I don't understand the comment - I think you might be looking at the old code instead of new (AX25 header doesn't have bitrate anymore). If that's not the case, let me know since I'm missing something.

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
anuraaga: From what I understand, there are communication protocols which define both framing and modulation as part of the standard (e.g., APT). In that case I think it's appropriate to keep them together, and later we'll have cases where modulation isn't necessarily provided since it's implied by the framing.

rtubio: First of all, sorry for the long reply, but I think that it is required that I properly explained all the things that are mixed in this discussion.

NOAA weather satellites are the main users of this APT protocol (https://en.wikipedia.org/wiki/Automatic_picture_transmission) although others like METEOR also implement it. The implementation for their satellites uses a specific modulation and bitrate, but this is just the implementation that they have chosen. Since nobody else was doing anything similar, people adopted these modulations and bitrates. However, you can use the APT framing format with a different set of modulations and bitrates; something that it is not done since the satellites and the ground equipment is now compatible with that whole data set.

This is something that happens sometimes in the telecommunications industry: certain protocols are used only with a set of underlaying radio devices because their application is very specific. In the end, the whole set kind of becomes a standard.

In our case, we should separate protocol from modulations and bitrates as a general rule. For example, AX.25 can be used with a broad range of modulations and bitrates.

For special cases like APT, I recommend labeling it as "APT/NOAA" or something alike; but always referring to the fact that we are using NOAA convention for the modulation, bitrate, polarization and central frequency. In this case, if the user selects the "APT/NOAA" protocol in the communications API, the parameters of the RadioConfiguration device should be set accordingly and users cannot modify them.

We might set a different protocol called "APT" where we use the framing as defined by NOAA, but we freely allow users to freely change the RadioConfiguration device parameters.

In addition to this, please also note that CW is a modulation and not a protocol (please refer to: https://en.wikipedia.org/wiki/Continuous_wave). On top of it, for beaconing from satellites, there is a data format that is widely used specially within the CubeSat community to just broadcast the codename for the satellite together with the battery voltage. But this is done on top of CW, that it is a modulation.

For example, in the case of the Birds satellites, their message in the beacon follows this format: 

http://birds2.birds-project.com/operation/

You can tell that they are using the same modulation (CW), but the frame format differs from others like: 

http://www.dk3wn.info/p/?cat=90

This was the beacon for our first satellite in Vigo called XaTcobeo (re-entered the atmosphere years ago).

Hope this helps - let me know if you wanna further discuss about this in person, it is easier most of the times! :)

anuraaga: I think this means modulation is independent from protocol so it should be separate. That makes sense, actually it means I don't need a BITSTREAM type of protocol anymore since the lack of a protocol is BITSTREAM.

rtubio: LGTM

PR #49, opened 2018-08-22
Improve C++ stub publishing

gavinreaney:
Note: cmake is required on the host machine to do package creation and publishing.

1) Bintray user and API key are read from gradle properties
2) CMakeLists use glob to find source files.
3) Only publish non-snapshot versions
4) Version number is automatically populated into conanfile.py via templating

### DIFF
--- a/stubs/cpp/build.gradle
+++ b/stubs/cpp/build.gradle
@@ -52,11 +52,28 @@ protobuf {
 
 }
 
-task downloadConanDeps {
-    dependsOn ':installPythonTools'
+task prepareConanfilePy {
+    inputs.property("version", version);
+    inputs.file("package/conanfile.py.tmpl")
+    outputs.file("$buildDir/generated/scripts/conanfile.py")
 
-    inputs.file 'conanfile.py'
-    outputs.dir 'build/conan/bin'
+    doFirst {
+        def template = file('package/conanfile.py.tmpl').text
+        def filled = template.replaceFirst('\\|API_VERSION\\|', version.toString())
+        def outDir = "$buildDir/generated/scripts"
+        mkdir outDir
+        def out = file("${outDir}/conanfile.py")
+        out.text = filled
+        out.executable = true
+    }
+}
+
+task downloadConanDeps {
+    dependsOn prepareConanfilePy, ':installPythonTools'
+
+    inputs.file "$buildDir/generated/scripts/conanfile.py"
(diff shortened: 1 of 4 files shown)


stubs/cpp/build.gradle
```diff
@@ -52,26 +52,49 @@ protobuf {
 
 }
 
+task prepareConanfilePy {
+    inputs.property("version", version);
```
anuraaga: Let's use single quotes for non-template strings.

stubs/cpp/build.gradle
```diff
@@ -52,26 +52,49 @@ protobuf {
...
+    outputs.file("$buildDir/generated/scripts/conanfile.py")
+
+    doFirst {
+        def template = file('package/conanfile.py.tmpl').text
+        def filled = template.replaceFirst('\\|API_VERSION\\|', version.toString())
+        def outDir = "$buildDir/generated/scripts"
+        mkdir outDir
+        def out = file("${outDir}/conanfile.py")
+        out.text = filled
+        out.executable = true
```
anuraaga: Don't think conanfiles need to be executable

stubs/cpp/build.gradle
```diff
@@ -84,4 +107,33 @@ afterEvaluate {
     tasks.generateProto.doLast {
         delete 'build/conan/src/google'
     }
+
+    def conanDir = 'build/conan'
+    def reference = 'stellarstation/testing'
```
anuraaga: Let's go ahead and use stable

stubs/cpp/build.gradle
```diff
@@ -84,4 +106,33 @@ afterEvaluate {
...
+        dependsOn createPackage
+
+        doLast {
+            exec {
+                def user = rootProject.property('bintray.user')
+                def key = rootProject.property('bintray.key')
+
+                executable 'bash'
+                args '-c', condaCommand(
+                        "conan user -p $key -r stellarstation $user && " +
```
anuraaga: I can't find this in my bash history or even setting environment variables, but I think we can avoid two steps by using variables

https://docs.conan.io/en/latest/reference/env_vars.html

gavinreaney: I switched to using env vars, but it seems to end up running 'conan user' anyway. (Conan detects the failure to authenticate, then gets the details from the env and runs conan user).

PR #51, opened 2018-08-23
Publish all stubs

gavinreaney:
Not sure if the 'onlyIf' checks are the best approach here since the behavior of 'publish' might be a bit mysterious.

One issue that people might run into is if the publishing partially succeeds - in that case re-running might be problematic if the repo doesn't allow skipping identical versions (PyPi does allow skipping if the version number is the same).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,6 +40,10 @@ allprojects {
         def bintrayUser = findProperty('bintray.user')
         def bintrayKey = findProperty('bintray.key')
 
+        // Skip the task if this is a snapshot.
+        tasks.bintrayUpload.onlyIf { !version.endsWith('SNAPSHOT') }
+        rootProject.tasks.bintrayPublish.onlyIf { !version.endsWith('SNAPSHOT') }
+
         afterEvaluate {
             bintray {
                 publish = true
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -40,6 +40,10 @@ allprojects {
         def bintrayUser = findProperty('bintray.user')
         def bintrayKey = findProperty('bintray.key')
 
+        // Skip the task if this is a snapshot.
```
anuraaga: We can have this but we don't need it either, the task will just fail if it's snapshot.

build.gradle
```diff
@@ -122,3 +126,10 @@ task installPythonTools(type: Exec) {
     executable 'bash'
     args '-c', condaCommand('pip install conan grpcio-tools twine')
 }
+
+task publishStubs {
```
anuraaga: How about adding a comment listing the gradle properties that need to be set

build.gradle
```diff
@@ -122,3 +122,17 @@ task installPythonTools(type: Exec) {
     executable 'bash'
     args '-c', condaCommand('pip install conan grpcio-tools twine')
 }
+
+// To use this task the following properites should be set in your gradle.properties:
+//   pypiUser
```
anuraaga: Shall we change the properties for consistency to `pypi.user` and `pypi.password`?

gavinreaney: Ah yeah. Will do.

PR #53, opened 2018-08-27
Add definition for ReservePass.

anuraaga:


api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -228,3 +254,54 @@ message Pass {
   // this pass.
   double uplink_center_frequency_hz = 10;
 }
+
+// A plan that specifies antenna configuration for tracking a satellite. A plan
```
kenichi-fukushima: Is it appropriate to say `antenna configuration` in documentation of an API for satellite operators?

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -228,3 +254,54 @@ message Pass {
...
+// is most often created from a pass but does not have to. A plan can contain
```
kenichi-fukushima: This second sentence is not appropriate for an external API.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -228,3 +254,54 @@ message Pass {
...
+    // The plan is currently executing and will end in the future at `los_time`.
+    EXECUTING = 1;
+
+    // The plan already finished executing successfully.
+    SUCCEEDED = 2;
+
+    // The plan failed to execute correctly.
+    FAILED = 3;
+
+    // The plan was canceled.
```
kenichi-fukushima: maybe mention that a plan can be canceled for a reason that the satellite operator cannot control.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -228,3 +254,54 @@ message Pass {
   // this pass.
   double uplink_center_frequency_hz = 10;
 }
+
+// A plan that specifies antenna configuration for tracking a satellite. A plan
+// is most often created from a pass but does not have to. A plan can contain
+// arbitrary antenna angles and communication frequencies.
+message Plan {
```
kenichi-fukushima: no need to include ground station information?

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
anuraaga: Thanks, completely forgot to update this comment.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
anuraaga: It wasn't in the spec, but it may be important. @hoshir can you confirm?

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
anuraaga: Yeah it's good to add a comment now. We're thinking of adding a field later with the actual reason included too.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -67,6 +67,20 @@ service StellarStationService {
   // If the satellite is not found or the client is not authorized for it, the stream will be closed
   // with a `NOT_FOUND` error.
   rpc OpenSatelliteStream (stream SatelliteStreamRequest) returns (stream SatelliteStreamResponse);
+
+  // Reserves a pass for later execution as a plan. The pass must be specified by a
+  // `reservation_token` returned in another method like `ListUpcomingAvailablePasses` and must be
```
gavinreaney: 'and must be reservable' made me think that we're not allowed to call the method until we've checked whether the pass is reservable. Not sure if other people would read it that way though.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -228,3 +254,53 @@ message Pass {
...
+
+  // The max elevation during the plan, in degrees.
+  double max_elevation_degrees = 6;
+
+  // The time of the max elevation during the plan.
+  google.protobuf.Timestamp max_elevation_time = 7;
+
+  // The center frequency, in Hz, for downlinking in this plan. 0 if downlink is not available in
+  // this plan.
+  double downlink_center_frequency_hz = 8;
```
hoshir: uint64 
Better to be consistent with RadioDeviceConfiguration.center_frequency_hz

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -228,3 +254,53 @@ message Pass {
...
+
+  // The center frequency, in Hz, for uplinking in this plan. 0 if uplink is not available in
+  // this plan.
+  double uplink_center_frequency_hz = 9;
```
hoshir: uint64

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -228,3 +254,53 @@ message Pass {
   // this pass.
   double uplink_center_frequency_hz = 10;
 }
+
+// A scheduled plan, specifying a time range within which a satellite will be communicated with.
```
hoshir: Since we call a scheduled pass as a Plan, "Scheduled plan" sounds a bit weird.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -228,3 +254,54 @@ message Pass {
...
+// arbitrary antenna angles and communication frequencies.
+message Plan {
+  // The unique ID of the plan.
+  string id = 1;
+
+  // The id of the satellite to be tracked in the plan.
+  string satellite_id = 2;
+
+  // A description of the status of a plan.
+  enum Status {
```
gavinreaney: As I understand it, the new API can only ever return 'RESERVED' here for the API we've added. Is that accurate?

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
hoshir: Correct, without the information, users won't be able to see GS information of the plan. I added it to the spec.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
gavinreaney: I'm not very excited about it but my attempt to make it clearer isn't going so well :) I have the following text. Basically, it's a bit messy that the same error code is used for different error types (AOS too close, schedule conflict).

// Reserves a pass for later execution as a plan. The pass must be specified by a
// `reservation_token` returned in another method like `ListUpcomingAvailablePasses`.
//
// If the `reservation_token` is invalid, the request will return an `INVALID_ARGUMENT` error.
//
// If the requested pass cannot be scheduled due to a conflict with existing plans or other
// scheduling constraints, the request will return a `FAILED_PRECONDITION` error. This includes
// if the pass `aos_time` is less than one hour in the future.
//
// Note that the reservation_token may have become unreservable since it was fetched. If that's the case,
// the call to ReservePass will fail with a 'FAILED_PRECONDITION' error and `ListUpcomingAvailablePasses`
// should be called again to get a fresh list of reservable passes. The pass must be scheduled at least
// one hour in advance of `aos_time` or it will be rejected.
//

PR #54, opened 2018-08-27
Build and push golang API stubs

gavinreaney:
Example commit and push when I added a field to a proto:
https://github.com/gavinreaney/go-stellarstation/commit/e3230c5b31934925235c8600ff2d770c94b389de

This will publish even for 'snapshot' versions.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,6 +16,8 @@
 
 plugins {
     id 'org.curioswitch.gradle-curiostack-plugin'
+    id 'com.github.blindpirate.gogradle' version '0.10' apply false
+    id 'org.ajoberstar.git-publish' version '1.0.1' apply false
 }
 
 gcloud {
--- a/build.gradle
+++ b/build.gradle
@@ -135,6 +135,7 @@ task publishStubs {
     description 'Publishes API stubs (only works for non-snapshot versions)'
     dependsOn ':api:bintrayUpload'
     dependsOn ':stubs:cpp:uploadPackage'
+    dependsOn ":stubs:golang:gitPublishCommit"
     dependsOn ':stubs:nodejs:publish'
     dependsOn ':stubs:python:uploadPythonPackage'
 }
--- a/gradle.properties
+++ b/gradle.properties
@@ -16,3 +16,7 @@
 
 org.curioswitch.curiostack.version = 0.0.143
 version = 0.0.9-SNAPSHOT
+
+# Avoid conflict of task name between gogradle and others
+# https://github.com/gogradle/gogradle/blob/master/docs/tasks.md#task-name-conflict
+org.gradle.jvmargs=-Dgogradle.alias=true
--- a/settings.gradle
+++ b/settings.gradle
@@ -39,5 +39,6 @@ include ':api'
 include ':examples:fakeserver'
 include ':examples:java:printing-client'
 include ':stubs:cpp'
+include ":stubs:golang"
 include ':stubs:nodejs'
 include ':stubs:python'


stubs/golang/build.gradle
```diff
@@ -0,0 +1,142 @@
...
+buildscript {
+    repositories {
+        mavenCentral()
+    }
+    dependencies {
+        classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.6'
+    }
+}
+
+apply plugin: 'com.github.blindpirate.gogradle'
```
anuraaga: I think you can do

```
plugins {
  id 'com.github.blindpirate.gogradle'
  id 'java'
  ...
}
```

stubs/golang/build.gradle
```diff
@@ -0,0 +1,142 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+buildscript {
+    repositories {
+        mavenCentral()
+    }
+    dependencies {
+        classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.6'
```
anuraaga: Can be in a different PR but your change above reminded me it'd be good to move the version for this plugin to the top level too.

stubs/golang/build.gradle
```diff
@@ -0,0 +1,142 @@
...
+    // resolution happens here so we insert our hack after it.
+    generateProto.actions.addAll(1, hackProtoWrap)
+
+    generateProto.doFirst {
+        mkdir file('build/proto-src')
+    }
+    generateProto.doLast {
+        copy {
+            from file('build/generated/source/proto/main/go/stellarstation')
+            into projectDir
```
anuraaga: We can copy into a directory under `build` now since this is published, not the repo itself. Also you may just be able to generate straight into the repo, though if it adds complication of us having to manually delete `api` or something, don't worry about it.

https://github.com/ajoberstar/gradle-git-publish#avoiding-extra-copy

stubs/golang/build.gradle
```diff
@@ -0,0 +1,144 @@
...
+    }
+    generateProto.doLast {
+        copy {
+            from file('build/generated/source/proto/main/go/stellarstation')
+            into gitPublish.repoDir
+        }
+    }
+}
+
+tasks.clean.doFirst {
```
anuraaga: Remove I think

stubs/golang/build.gradle
```diff
@@ -0,0 +1,144 @@
...
+    repoUri = 'git@github.com:infostellarinc/go-stellarstation.git'
+    branch = 'master'
+
+    preserve {
+        include '**'
+        exclude 'api'
+    }
+
+    // what to publish, this is a standard CopySpec
+    contents {
```
anuraaga: Do you still need this even though we generate into the repo?

build.gradle
```diff
@@ -16,6 +16,8 @@
 
 plugins {
     id 'org.curioswitch.gradle-curiostack-plugin'
+    id 'com.github.blindpirate.gogradle' version '0.10' apply false
```
hoshir: Alphabetical order

stubs/golang/build.gradle
```diff
@@ -0,0 +1,134 @@
...
+    repositories {
+        mavenCentral()
+    }
+    dependencies {
+        classpath 'com.google.protobuf:protobuf-gradle-plugin:0.8.6'
+    }
+}
+
+plugins {
+    id 'com.github.blindpirate.gogradle'
```
hoshir: Order by name.

gavinreaney:
go get was failing because git is not found by the circleci build.

I'm not sure of the best fix, but updating the circleci image to circleci/openjdk should theoretically work but it failed restoring the cache. This is likely because the official circleci builds run as a normal user rather than root, so can't access the cache directories.

I will try using a new cache key to see if that works.

anuraaga:
I guess circleci images may not need to be pulled so it seems like a good change either way. Surprised you don't need to add a tag for jdk10 though, do you know what jdk it's using?

stubs/golang/build.gradle
```diff
@@ -0,0 +1,134 @@
...
+    id 'java'
+    id 'com.google.protobuf'
+    id 'org.ajoberstar.git-publish'
+}
+
+golang {
+    packagePath = 'github.com/infostellarinc/go-genproto'
```
hoshir: Will this be used as a package name of stubs for Go lang? If so, it should be consistent with GitHub repository to push. 

Since go_package is defined in each proto file, I am not sure if this is needed.
https://github.com/infostellarinc/stellarstation-api/commit/7e8631ce4b4a7e9fda2139f0554bbe120f6713ca

hoshir: Interesting, thanks for the explanation.

stubs/golang/build.gradle
gavinreaney: In this case java needs to be before the protobuf application otherwise the build fails (it assumes java stuff is already present).

build.gradle
```diff
@@ -15,9 +15,9 @@
  */
 
 plugins {
-    id 'org.curioswitch.gradle-curiostack-plugin'
     id 'com.github.blindpirate.gogradle' version '0.10' apply false
     id 'org.ajoberstar.git-publish' version '1.0.1' apply false
+    id 'org.curioswitch.gradle-curiostack-plugin'
```
anuraaga: I'm glad I make my plugin order agnostic :)

PR #57, opened 2018-09-03
Cloudbuild config to push snapshots

gavinreaney:
I had a lot of trouble with the git authentication but this seems to work. I haven't yet pushed a newer version of the go stubs (but when I tested locally, the diff looked sane).


cloudbuild/snapshot-publish.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - --key=deploy-stubs
+- name: openjdk:10-jdk
+  entrypoint: bash
+  args:
+  - -c
+  # The SSH clients are looking in /root/.ssh for keys and config, even although the HOME directory
+  # is /builder/home.
+  - |
+    mkdir /root/.ssh &&
+    cp cloudbuild/id_rsa /root/.ssh/id_rsa &&
```
anuraaga: What do you think about using a volume for `/root/.ssh`? It only saves these two lines though.

gavinreaney:
I didn't get around to adding caching stuff yet. Could maybe look at that later? Not sure how often we'll be pushing these snapshots.

cloudbuild/snapshot-publish.yaml
gavinreaney: I was a bit paranoid about having secrets in a volume although I guess it isn't so different from them being in the container directly. I'll probably just keep it as is since it works.

PR #63, opened 2018-09-13
STELLAR-1630: Add telemetry data to Plans response proto

gavinreaney:


api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -316,6 +316,26 @@ message Pass {
   uint64 uplink_center_frequency_hz = 10;
 }
 
+// Data uploaded from the ground station during a pass.
+message TelemetryData {
+  // URL of the data. This URL is valid for one hour after being returned.
+  string url = 1;
+
+  // The type of data available at the above URL.
+  enum DataType {
```
anuraaga: I believe this is the same as `FramingType` - RAW == IQ, DEMODULATED == BITSTREAM, DECODED == AX.25, etc.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -316,6 +316,26 @@ message Pass {
...
+
+    // Demodulated IQ data.
+    DEMODULATED = 1;
+
+    // Decoded data.
+    DECODED = 2;
+  }
+  // The status of this plan.
+  DataType data_type = 2;
+}
```
anuraaga: I think we'll want to return zip archives of many decoded packets in some cases - do you think we'll later add an enum of archivetype?

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -316,6 +316,26 @@ message Pass {
   uint64 uplink_center_frequency_hz = 10;
 }
 
+// Data uploaded from the ground station during a pass.
```
kenichi-fukushima: The term `uploaded` sounds like it puts too much emphasis on the server-client architecture of StellarStation as a software service. Should we use a terminology of satellite communication such as `downlinked`?

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
gavinreaney: @hoshir can you comment on this?

I think I understand that RAW == IQ and DEMODULATED == BITSTREAM but I thought DECODED was more like getting an image, text file or whatever.

gavinreaney: After talking to Andrew I think I get it now. Decoded means that we've removed the framing data (e.g. AX.25 headers etc).

I think it would be confusing to mention framing type here.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
gavinreaney: I'm hoping that the storage URL we return can return a content-type header when it's fetched so we don't need to include specific data types here.

anuraaga: Having to parse a content-type will be much harder to use I think. But either way, it should be simple enough to add later.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
hoshir: The enum values we want to use match with Framing enum, BITSTREAM, AX25, IQ (and IMG_PNG). However, using the enum to a variable named data_type sounds confusing to me. For me, changing variable name to something like "Framing data_framing" sounds natural when we reuse Framing enum.

gavinreaney: I noticed the framing enum also includes IMAGE_PNG and is used in requests too.

If we can freely add additional values for other decoded data types then it is convenient to reuse it. The name 'framing' feels not very accurate though.

I'll consider updating the API later.

PR #64, opened 2018-09-13
Add definition for returning stream events.

anuraaga:


api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -125,6 +125,15 @@ message SatelliteStreamRequest {
...
   string stream_id = 2;
 
+  // A unique ID identifying this request. If filled, monitoring events will be returned in
+  // `SatelliteStreamResponse.stream_event` with this value copied into `StreamEvent.request_id` for
+  // correlation. The API will treat this as an opaque identifier which can be set to any value,
+  // e.g., to the same value as the request ID in a downstream system. For simple cases, it is often
+  // simplest to set this to a standard UUID.
+  //
+  // If unset, stream events wil not be returned.
+  string request_id = 5;
```
hoshir: What's the reason of returning StreamEvent only when requieest_id is filled? I guess it's for performance, but want to confirm it.

anuraaga: Yeah, disabling events if not needed can squeeze some extra performance, and the value of events is much lower without the ability to relate to a request so I'd like it to be a condition for having events.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -217,6 +230,29 @@ message Telemetry {
   bytes frame_header = 6;
 }
 
+// A event that occurred while processing the stream. Only returned if `request_id` was populated in
```
gavinreaney: An event

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
hoshir: What do you think of the idea that add "bool enable_monitoring" and return the event when it is true? I proposed this idea because asking user to set request_id is somewhat unusual, and guess it might be convenient for them we add the id, though the idea may not work if order of response is not guaranteed.

anuraaga: Yeah it's mainly because order isn't totally guaranteed and it's all asynchronous. I think setting a UUID is very easy in all the languages (except maybe C++...), and we should have it example code we publish. What do you think?

hoshir: OK, we can update it when we have better idea.

kenichi-fukushima: - Will there be cases where there is an event that relates to no request or more than one request?
- Do we allow that the first request has id but subsequent ones do not?

anuraaga: Thanks - thinking about the case of events with no request made me realize having a separate boolean makes sense. I don't think there is a case for events with multiple requests, so will keep that singular.

Also, since this is just an opaque passthrough, I don't think we need to over-specify it. I went ahead and documented in detail about the concept of the "first request" on a stream, which is useful for our other parameters like `satellite_id` anyways. Let me know what you think about the new docs.

PR #70, opened 2018-10-15
Add basic support for unavailability windows to the external API.

gavinreaney:
I wasn't overly specific in the documentation about when a plan cannot be canceled.

### DIFF
--- a/api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
+++ b/api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
@@ -64,8 +64,8 @@ service GroundStationService {
 
   // Returns a list of unavailability windows for the requested ground station.
   //
-  // The request will be closed with an `INVALID_ARGUMENT` status if `ground_station_id` is missing
-  // or invalid.
+  // The request will be closed with an `INVALID_ARGUMENT` status if `ground_station_id`,
+  // `start_time`, or `end_time` are missing, or 'end_time' is not after 'start_time'.
   rpc ListUnavailabilityWindows (ListUnavailabilityWindowsRequest) returns (ListUnavailabilityWindowsResponse);
 }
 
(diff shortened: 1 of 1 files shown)


api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -107,3 +128,55 @@ message Tle {
   // The second line of the TLE.
   string line_2 = 2;
 }
+
+// A time window during which a ground station is unavailable e.g. for local maintenance.
+message UnavailabilityWindow {
```
kenichi-fukushima: It is very likely we have recurring windows and satellite-based windows. Should we design these data structures with those extensions in mind from the beginning?

gavinreaney: Thanks for raising this point.

I got some feedback from others that clients will use results from pass computation to manually create windows instead of us supporting satellite based in the API (either the caller has their own pass computer or we'll make an API for that).

Also, since it isn't too much effort to write code for generating a recurring window from this API, I didn't add that.

But having said the above, I can't rule out supporting those in future so I'll look into making the change.

kenichi-fukushima: The minimalist API approach sounds fine to me if it is a realistic assumption that an average customer is computer-literate enough to hit assemble and send API requests.

gavinreaney: I was thinking about this in more detail.

I think the request and response need to modeled slightly differently from each other. E.G. in the response, the user probably wants to have the recurring window expanded (certainly they'd want this for satellite based). I think that means the start and end time are necessary for those, and we'd just add extra fields for the other stuff.

On the request side a oneof does seem more reasonable but then again I don't think it is too much trouble to document that the start and end time don't apply for satellite based (f we support it in future).

So I'm slightly favoring just keeping it as-is.

anuraaga: Since we're not using this message in Add that sounds reasonable to me too.

PR #74, opened 2018-10-25
Add enums for modulation / coding parameters to use in metrics.

anuraaga:
Please help me write good documentation for these fields - I sort of researched / assumed somethings but doubt they are ideal. Let's err on the side of detail vs not - I like to think of a metric for our API documentation as it's good if it teaches some of the basics of DSP.

Also, if anything applies only to reception and not transmission, we should document such semantics too.

### DIFF
--- a/api/src/main/proto/stellarstation/api/v1/radio/radio.proto
+++ b/api/src/main/proto/stellarstation/api/v1/radio/radio.proto
@@ -89,4 +89,174 @@ enum Modulation {
 
   // Binary Phase Shift Keying. https://en.wikipedia.org/wiki/Phase-shift_keying#Binary_phase-shift_keying_(BPSK)
   BPSK = 2;
+
+  // Multiple Frequency Shift Keying. https://en.wikipedia.org/wiki/Multiple_frequency-shift_keying
+  MFSK = 3;
+
+  // Quadrature Phase Shift Keying. https://en.wikipedia.org/wiki/Phase-shift_keying#Quadrature_phase-shift_keying_(QPSK)
+  QPSK = 4;
+
+  // 8-Phase Shift Keying. https://en.wikipedia.org/wiki/Phase-shift_keying
+  PSK8 = 5;
+
+  // 16-Phase Shift Keying. https://en.wikipedia.org/wiki/Phase-shift_keying
+  PSK16 = 6;
+
+  // 32-Phase Shift Keying. https://en.wikipedia.org/wiki/Phase-shift_keying
+  PSK32 = 7;
+
+  // 64-Phase Shift Keying. https://en.wikipedia.org/wiki/Phase-shift_keying
+  PSK64 = 8;
+
+  // 128-Phase Shift Keying. https://en.wikipedia.org/wiki/Phase-shift_keying
+  PSK128 = 9;
+
+  // 256-Phase Shift Keying. https://en.wikipedia.org/wiki/Phase-shift_keying
+  PSK256 = 10;
+
+
+  // Differential Phase Shift Keying. https://en.wikipedia.org/wiki/Phase-shift_keying#Differential_phase-shift_keying_(DPSK)
+  DPSK = 11;
+
+  // Optical Differential Phase Shift Keying. https://www.bell-labs.com/our-research/publications/203155/
+  ODPSK = 12;
+
+  // Offset Quadrature Phase Shift Keying. https://en.wikipedia.org/wiki/Phase-shift_keying#Offset_QPSK_(OQPSK)
+  OQPSK = 13;
+
+  // Asynchronous Quadrature Phase Shift Keying.
+  AQPSK = 14;
+
+
+  // 8-Phase Amplitude Modulation. https://en.wikipedia.org/wiki/Quadrature_amplitude_modulation
+  QAM8 = 15;
+
+  // 16-Phase Amplitude Modulation. https://en.wikipedia.org/wiki/Quadrature_amplitude_modulation
+  QAM16 = 16;
+
+  // 32-Phase Amplitude Modulation. https://en.wikipedia.org/wiki/Quadrature_amplitude_modulation
+  QAM32 = 17;
+
+  // 64-Phase Amplitude Modulation. https://en.wikipedia.org/wiki/Quadrature_amplitude_modulation
+  QAM64 = 18;
+
+  // 128-Phase Amplitude Modulation. https://en.wikipedia.org/wiki/Quadrature_amplitude_modulation
+  QAM128 = 19;
+
+  // 256-Phase Amplitude Modulation. https://en.wikipedia.org/wiki/Quadrature_amplitude_modulation
+  QAM256 = 20;
+
+
+  // Minimum Shift Keying. https://en.wikipedia.org/wiki/Minimum-shift_keying
+  MSK = 21;
+
+  // Gaussian Minimum Shift Keying. https://en.wikipedia.org/wiki/Minimum-shift_keying#Gaussian_minimum-shift_keying
+  GMSK = 22;
+
+
+  // Amplitude modulation (generic)
+  AM = 23;
+
+  // Frequency modulation (generic)
+  FM = 24;
+
+  // Phase modulation (generic)
+  PM = 25;
+
+  // Pulse-code modulation (generic)
+  PCM = 26;
+}
+
+// Line coding applied during modulation. Line coding is the technique for modeling binary 0 and 1
+// within a modulated waveform.
+enum LineCoding {
+  // Non-return-to-zero level. https://en.wikipedia.org/wiki/Non-return-to-zero
+  NRZ_L = 0;
+
+  // Non-return-to-zero mark. https://en.wikipedia.org/wiki/Non-return-to-zero
+  NRZ_M = 1;
+
+  // Non-return-to-zero space. https://en.wikipedia.org/wiki/Non-return-to-zero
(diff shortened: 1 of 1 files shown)


api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -89,4 +89,174 @@ enum Modulation {
...
+  PSK128 = 9;
+
+  // 256-Phase Shift Keying. https://en.wikipedia.org/wiki/Phase-shift_keying
+  PSK256 = 10;
+
+
+  // Differential Phase Shift Keying. https://en.wikipedia.org/wiki/Phase-shift_keying#Differential_phase-shift_keying_(DPSK)
+  DPSK = 11;
+
+  // Optical Differential Phase Shift Keying. https://www.bell-labs.com/our-research/publications/203155/
```
vaelen: I was thinking "Offset DPSK" here, but I can't find it anywhere now, so let's drop it.

anuraaga:
I have written so many protos in my life yet haven't run into colliding ENUM names before like here. @kenichi-fukushima any thoughts on a pattern of wrapping enums with messages just for namespacing?

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -89,4 +89,174 @@ enum Modulation {
...
+  ODPSK = 12;
+
+  // Offset Quadrature Phase Shift Keying. https://en.wikipedia.org/wiki/Phase-shift_keying#Offset_QPSK_(OQPSK)
+  OQPSK = 13;
+
+  // Asynchronous Quadrature Phase Shift Keying.
+  AQPSK = 14;
```
vaelen: This one seems obscure enough that we should probably not include it.  It seems to only be used in cellphone systems.

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -89,4 +89,174 @@ enum Modulation {
...
+  // No Viterbi coding.
+  OFF = 0;
+
+  // Only Viterbi descrambling.
+  DESCRAMBLING = 1;
+
+  // Only Viterbi decoding.
+  DECODING = 2;
+
+  // Only inverted Viterbi.
```
vaelen: "Only inverted Vitberbi decoding."

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -89,4 +89,174 @@ enum Modulation {
...
+  BP_L = 4;
+
+  // Bipolar return-to-zero mark. https://en.wikipedia.org/wiki/Bipolar_encoding
+  BP_M = 5;
+
+  // Bipolar return-to-zero space. https://en.wikipedia.org/wiki/Bipolar_encoding
+  BP_S = 6;
+}
+
+// A type of Viterbi coding applied to a bitstream.
```
vaelen: https://en.wikipedia.org/wiki/Convolutional_code

vaelen: "Viterbi convolutional coding"

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -89,4 +89,174 @@ enum Modulation {
...
+  DECODING_INVERTED = 3;
+
+  // Viterbi decoding and descrambling.
+  DECODING_DESCRAMBLING = 4;
+
+  // Inverted Viterbi decoding and descrambling.
+  DECODING_INVERTED_DESCRAMBLING = 5;
+}
+
+// A type of Reed-Solomon coding applied to a bitstream.
```
vaelen: https://en.wikipedia.org/wiki/Reed%E2%80%93Solomon_error_correction

vaelen: "Reed-Solomon error correction"

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -89,4 +89,174 @@ enum Modulation {
...
+enum ReedSolomonMode {
+  // No Reed-Solomon coding.
+  OFF = 0;
+
+  // Alpha Reed-Solomon coding.
+  ALPHA = 1;
```
vaelen: I'm not actually sure what Alpha vs Beta means here.  I wonder if for our reporting purposes this should just be ON or OFF.  @reiinakano , what do you think?

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -89,4 +89,174 @@ enum Modulation {
...
+
+  // Beta Reed-Solomon coding.
+  BETA = 2;
+}
+
+// A type of frame checking applied to a bitstream.
```
vaelen: "frame error checking"

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -89,4 +89,174 @@ enum Modulation {
...
+  OFF = 0;
+
+  // Cyclic redundancy check.
+  CRC = 1;
+
+  // Generic checksum.
+  CHECKSUM = 2;
+}
+
+// A current status of Viterbi coding.
```
vaelen: "Viterbi convolutional coding"

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -89,4 +89,174 @@ enum Modulation {
...
+
+  // 128-Phase Shift Keying. https://en.wikipedia.org/wiki/Phase-shift_keying
+  PSK128 = 9;
+
+  // 256-Phase Shift Keying. https://en.wikipedia.org/wiki/Phase-shift_keying
+  PSK256 = 10;
+
+
+  // Differential Phase Shift Keying. https://en.wikipedia.org/wiki/Phase-shift_keying#Differential_phase-shift_keying_(DPSK)
+  DPSK = 11;
```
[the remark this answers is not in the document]
vaelen: Ahh, that's good to know.

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
reiinakano: Oh! For this one and other comments like it, just "convolutional coding" is more accurate. Viterbi decoding is a special type of "convolutional decoding" and these parameters actually have nothing to do with the Viterbi algorithm, but the underlying convolutoinal code.

kenichi-fukushima:
@anuraaga I think it is reasonable to have some structure to organizing enums with very concise names.

An alternative to enclosing messages is packages.

kenichi-fukushima:
Packages are far less convenient than messages in languages like Java that does not provide ways to defines aliases of packages, though.

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -89,4 +89,153 @@ enum Modulation {
...
+// Parameters for convolutional coding applied to a bitstream.
+// https://en.wikipedia.org/wiki/Convolutional_code
+message ConvolutionalCodingParams {
+  // The constraint length, `k`, of the code.
+  uint32 k = 1;
+
+  // The inverse of the rate of coding, e.g., if rate is `1 / r`, `inverse_rate` is `r`.
+  uint32 inverse_rate = 2;
+
+  message DescramblingParams {
```
reiinakano: Sorry, should've been clearer. Descrambling params should be on the same level as convolutional coding params since they're not related.

convolutional coding params:
 - k
 - inverse rate
 - polynomials (`repeated bytes` where each of the r `bytes` has length 'k' bits)

descrambling params:
 - type (additive/multiplicative)
 - length
 - polynomial mask for shift register (single `bytes` of length `length` bits)
 - seed (single `bytes` of length `length` bits)

reiinakano:
NRZ_L and NRZ_M are the actual terms, so let's stick to that, it's just that it might be helpful for satellite operators to know that those map directly to non-diff and differential.

reiinakano:
On second thought, I'm just speaking for myself, maybe expert satellite operators will have no problem differentiating those :)

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
anuraaga: Thanks - sorry for not knowing anything about this stuff ;)

anuraaga:
@rtubio Can you also take a look at the names and comments in this PR?

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -195,32 +195,9 @@ message ConvolutionalCodingParams {
...
-    // The length of the shift register, in bits. The most significant `length` bits of `seed` will
-    // be used.
-    uint32 length = 4;
-  }
-  // Descrambling parameters used when applying convolutional coding. If unset, descrambling is not
-  // used.
-  DescramblingParams descrambling = 3;
+  // The polynomials used for coding. Must have `r` number of elements, where the most significant
+  // `k` bits of each element will be used as the polynomial for that index.
+  repeated bytes polonymial = 3;
```
reiinakano: typo

PR #77, opened 2018-10-31
Add messages for returning ground station configuration / state for m…

anuraaga:
…onitoring.

As always, many comments are somewhat vague since I don't have a detailed understanding so appreciate any detailed reviews. And do want to err on the side of over-detail, e.g., we should teach good monitoring practices through our documentation for users that have less familiarity with the subject.

Initially, the messages will be used in pass monitoring - configuration is returned in the stream at the beginning of a pass, and state is returned periodically during the pass. Configuration messages may or may not be reusable elsewhere, it's not yet clear to me (thinking the antenna one should be).


api/src/main/proto/stellarstation/api/v1/antenna/antenna.proto
```diff
@@ -0,0 +1,71 @@
...
+
+// Messages for modeling physical antenna configuration in API messages.
+
+// The static configuration of a antenna. These are the physical parameters of the antenna.
+message AntennaConfiguration {
+
+  // The latitude of the antenna, in degrees.
+  double latitude = 1;
+
+  // The latitude of the antenna, in degrees.
```
hoshir: longitude

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
```diff
@@ -0,0 +1,235 @@
...
+  ReedSolomonStatus reed_solomon_status = 9;
+}
+
+// The current state of the ground station's antenna during the operation of a pass.
+message AntennaState {
+
+  // The state of an antenna angle.
+  message Angle {
+    // The angle that was specified for the antenna. When the antenna is functioning properly,
+    //    // there should be nearly no difference between this and `measured`.
```
hoshir: Duplicated `//`

api/src/main/proto/stellarstation/api/v1/antenna/antenna.proto
```diff
@@ -0,0 +1,71 @@
...
+
+package stellarstation.api.v1.antenna;
+
+option go_package = "github.com/infostellarinc/go-stellarstation/api/v1/antenna";
+
+option java_multiple_files = true;
+option java_outer_classname = "AntennaProto";
+option java_package = "com.stellarstation.api.v1.antenna";
+
+// Messages for modeling physical antenna configuration in API messages.
```
gavinreaney: Sounds a bit weird with the word messages twice. Maybe just:

"Messages for modeling physical antenna configuration in the API."

api/src/main/proto/stellarstation/api/v1/antenna/antenna.proto
```diff
@@ -0,0 +1,71 @@
...
+
+// The static configuration of a antenna. These are the physical parameters of the antenna.
```
gavinreaney: an Antenna

api/src/main/proto/stellarstation/api/v1/antenna/antenna.proto
```diff
@@ -0,0 +1,71 @@
...
+message AntennaConfiguration {
+
+  // The latitude of the antenna, in degrees.
+  double latitude = 1;
+
+  // The longitude of the antenna, in degrees.
+  double longitude = 2;
```
gavinreaney: The naming isn't consistent with the others in the message which append the units. In other words, should this be:
longitude_degrees?

OTOH, it might be so standard to just use latitude and longitude that you could argue for the exception.

api/src/main/proto/stellarstation/api/v1/antenna/antenna.proto
```diff
@@ -0,0 +1,71 @@
...
+  // https://en.wikipedia.org/wiki/Euler_angles#Tait-Bryan_angles
+  double pitch_degrees = 5;
+
+  // The roll (i.e., circular rotation), in degrees.
+  // https://en.wikipedia.org/wiki/Euler_angles#Tait-Bryan_angles
+  double roll_degrees = 6;
+
+  // The beamwidth, in degrees.
+  // https://en.wikipedia.org/wiki/Beamwidth
+  double beamwidth = 7;
```
gavinreaney: Again, should _degrees be appended?

api/src/main/proto/stellarstation/api/v1/antenna/antenna.proto
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// The polarization of an antenna.
```
gavinreaney: Maybe add wikipedia link? https://en.wikipedia.org/wiki/Antenna_(radio)#Polarization

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
```diff
@@ -0,0 +1,235 @@
...
+
+option go_package = "github.com/infostellarinc/go-stellarstation/api/v1/monitoring";
+
+option java_multiple_files = true;
+option java_outer_classname = "MonitoringProto";
+option java_package = "com.stellarstation.api.v1.monitoring";
+
+// Messages for reporting information about a ground station for monitoring and debugging.
+
+// The current configuration of a transmitter. A transmitter is taking a bitstream and applies
```
gavinreaney: Maybe:

A transmitter takes a bitstream ...

I would consider starting a new paragraph with this sentence (it seems fairly common style to have a single sentence summary paragraph followed by more details in separate paragraphs).

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
```diff
@@ -0,0 +1,235 @@
...
+  radio.LineCoding line_coding = 4;
+
+  // The bandwidth of the filter.
+  uint64 filter_bandwidth_hz = 5;
+
+  // The sampling frequency.
+  uint64 sampling_frequency_hz = 6;
+}
+
+// The current configuration of a receiver. A receiver is taking a waveform from a satellite and
```
gavinreaney: A receiver is taking => A receiver takes ?

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
```diff
@@ -0,0 +1,235 @@
...
+  uint64 bitrate = 4;
+
+  // The line coding of the receiver.
+  radio.LineCoding line_coding = 5;
+
+  // The frame synchronization word being used. Unset if frame synchronization is disabled.
+  common.Bits synchronization_word = 6;
+
+  // Whether convolutional code descrambling is enabled.
+  bool is_convolutional_code_descrambling = 7;
```
gavinreaney: Are all of these booleans independent?

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
```diff
@@ -0,0 +1,235 @@
...
+  // should match the parameters of the executing pass. When debugging issues with data
+  // reception (e.g., can't decode signal), it can be useful to confirm these values match
+  // the expected configuration of the receiver.
+  ReceiverConfiguration receiver = 2;
+
+  // The current configuration of the antenna used by the ground station. These are the physical
+  // parameters of an antenna and in practice should never change, even across passes. But natural
+  // events such as earthquakes can cause small changes in these values. When debugging issues with
+  // tracking a satellite (e.g., getting no signal at all), it can be useful to confirm these values
+  // match the expected configuration of the antenna.
```
gavinreaney: Ah so heading, roll, pitch are fixed also?

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
```diff
@@ -0,0 +1,235 @@
...
+    CORRECTION_CAPACITY_EXCEEDED = 2;
+  }
+  // Current status of Reed-Solomon coding.
+  Status status = 1;
+
+  // The number of Reed-Solomon corrected signals. Only set is `status` == `OK`.
+  uint32 num_corrected_signals = 2;
+}
+
+message ReceiverState {
```
gavinreaney: Comment for the message

api/src/main/proto/stellarstation/api/v1/antenna/antenna.proto
```diff
@@ -0,0 +1,71 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+syntax = "proto3";
+
+package stellarstation.api.v1.antenna;
```
gavinreaney: One thing I wondered about. The transmitter and receiver configuration are inside the monitoring proto. What's the reason for not also putting this message in the same proto?

Or maybe receiver and transmitter config could be considered antenna related and could be here? Not sure :)

hoshir:
LGTM

kenichi-fukushima:
Can you provide context of this PR? Without knowing who uses the API for what purpose, I cannot review it.

anuraaga:
Added some more context to the description

api/src/main/proto/stellarstation/api/v1/antenna/antenna.proto
anuraaga: Yeah I had some trouble deciding what to try to make more generic and what's monitoring specific. Since an antenna is physical and can't really change, I figure it could be returned with passes, etc too while for radios, since they're configurable, in such cases we'd need to return something quite different listing "supported options".

api/src/main/proto/stellarstation/api/v1/antenna/antenna.proto
anuraaga: Yeah we've usually been considering lat/lng being degrees so standard as not to include units. I agree it seems a bit weird for only these two to be non-suffixed, but I guess it's ok.

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
```diff
@@ -0,0 +1,235 @@
...
+
+option java_multiple_files = true;
+option java_outer_classname = "MonitoringProto";
+option java_package = "com.stellarstation.api.v1.monitoring";
+
+// Messages for reporting information about a ground station for monitoring and debugging.
+
+// The current configuration of a transmitter. A transmitter is taking a bitstream and applies
+// signal processing to create the final waveform sent to a satellite.
+message TransmitterConfiguration {
```
reiinakano: A lot of these are CORTEX-specific (Same with `ReceiverConfiguration`). For a generic representation, I suggest we just use the existing `RadioDeviceConfiguration`. Then modify `RadioDeviceConfiguration` to have `line_coding`, `convolutional_coding_params` (can be disabled), `reed_solomon_params` (can be disabled), `scrambling_params` (can be disabled), `sync_word`.

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
anuraaga: I think inverted / decoding are linked, added a comment.

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
anuraaga: @vaelen Can you confirm? I think they're the fixed rotation of the base of the antenna, which is basically screwed into the Earth. Azimuth / elevation are the moving angles.

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
```diff
@@ -0,0 +1,235 @@
...
+
+  // Locked onto the quadrature signal.
+  Q_LOCKED = 2;
+
+  // Locked onto both in-phase and quadrature signal.
+  LOCKED = 3;
+}
+
+// The status of Reed-Solomon coding.
+message ReedSolomonStatus {
```
reiinakano: These statuses (also `ConvolutionalCodingStatus`) make sense but I'm worried they might be too CORTEX-specific. It might be difficult to find other radios that give the same information. Is it a good approach to hold off on CORTEX-specific things until we find another radio that provide the same information?

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
```diff
@@ -52,8 +54,10 @@ message TransmitterConfiguration {
   uint64 sampling_frequency_hz = 6;
 }
 
-// The current configuration of a receiver. A receiver is taking a waveform from a satellite and
-// applying signal processing to produce a bitstream.
+// The current configuration of a receiver.
+//
+// A receiver takes a waveform from a satellite and applying signal processing to produce a
```
gavinreaney: applies

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
vaelen: I'm actually not sure.  We need to ask the ground station vendor to clarify.

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
anuraaga: From what I understood, the the device monitoring API doesn't return many of these parameters, and mostly just booleans like I have here. Is there a different way for us to verify this information at the beginning of a pass? If so, I definitely would rather just use `RadioDeviceConfiguration`

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
anuraaga: State we can monitor will vary by ground station but I think it's fine to include the joint set in our monitoring APIs and fill what we can - the more information for monitoring the better. Though if there are any fields you think are just never going to be useful for monitoring let me know and we should remove those.

anuraaga:
@reiinakano I think instead of naming the fields, we can add additional details to the comments / documentation as we better understand what parameters mean, which ones are important, and which radios support what.

api/src/main/proto/stellarstation/api/v1/antenna/antenna.proto
```diff
@@ -0,0 +1,72 @@
...
+
+  // The longitude of the antenna, in degrees.
+  double longitude = 2;
+
+  // The altitude of the antenna, in meters.
+  double altitude_meters = 3;
+
+  // The heading (i.e., left-right rotation), in degrees.
+  // https://en.wikipedia.org/wiki/Euler_angles#Tait-Bryan_angles
+  double heading_degrees = 4;
```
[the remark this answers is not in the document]
anuraaga: @vaelen What do you think? I think we may just want to remove the rotation fields, I have a feeling they wouldn't be very standard across ground stations.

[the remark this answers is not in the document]
anuraaga: They are values that the radio's monitoring API returns and we're erring on including everything the radio returns in out monitoring API too. But if you think it's too out there we should remove it. What do you think?

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
rtubio: @anuraaga pitch, roll, heading (aka yaw) is used for airplanes and satellites because they have 3 free axes of rotation. A ground station has only 2 free axes and its pointing is described with AZ/EL.

api/src/main/proto/stellarstation/api/v1/antenna/antenna.proto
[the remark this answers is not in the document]
rtubio: ACU's can fine track a satellite using the information on the electromagnetic wave, regardless of the TLE files. This specific ACU is estimating the attitude of the satellite using the incoming wave.

Would you please mind referring me to the specific section of the manual where these numbers are defined?

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
rtubio: These three angles are the estimated yaw, pitch, roll of the satellite reported by the ACU. The ACU has an automatic mode to report these parameters.

Since there is no functional description of this **"automatic attitude mode"** (it only appears listed on page 15), @vaelen will request Zodiac guys the user manual of the ACU.

rtubio: @anuraaga we can remove this parameters from the API since: (1) not all supported modems / radios will have this option, (2) we do not know if the attitude of all satellites can be estimated with this mode.

api/src/main/proto/stellarstation/api/v1/antenna/antenna.proto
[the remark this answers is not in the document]
kenichi-fukushima: Just FYI: I found this when I was reading manual of another MCS.
https://drive.google.com/drive/u/0/folders/1_RuFaq6L42giRT5sws4rpZH1ueTgnUJ_

page 13 of Software ICD V7.1.pdf
> The <station> element has the following properties:
 name: The name of the station.
 latitudeDeg: The station latitude in degrees north (southerly latitudes are negative).
 longitudeDeg: The station longitude in degrees east (westerly longitudes are negative).
 altitudeM: The station altitude (above sea level) in meters.
 __headingDeg: The station 'heading' in degrees (the physical orientation of the pedestal).__
 (optional) isDefault: 1 if this station is the default station (selected when the session is
loaded), 0 otherwise

This software only reports heading from those three angles and I found the description too ambiguous.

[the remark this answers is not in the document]
rtubio: In accordance with the available documentation, these values are the estimation of the attitude of the satellite, when the MCS is configured to work in an automatic mode called "Attitude". @vaelen is going to check with the provider for documentation with specific definitions.

Anyhow, @anuraaga : my recommendation is that we remove these values from the API since they are very specific for this device.

api/src/main/proto/stellarstation/api/v1/antenna/antenna.proto
```diff
@@ -0,0 +1,72 @@
...
+// The static configuration of an antenna. These are the physical parameters of the antenna.
+message AntennaConfiguration {
+
+  // The latitude of the antenna, in degrees.
+  double latitude = 1;
+
+  // The longitude of the antenna, in degrees.
+  double longitude = 2;
+
+  // The altitude of the antenna, in meters.
```
YutaKomiya: Please note that altitude is above mean sea level. 
@rtubio Could you suggest me the necessity to note referenced world geodetic system model to the coordination when we control ACU for S/X-band directly based on trajectory calculated by StellarStation?

api/src/main/proto/stellarstation/api/v1/antenna/antenna.proto
```diff
@@ -0,0 +1,72 @@
...
+
+  // The longitude of the antenna, in degrees.
+  double longitude = 2;
+
+  // The altitude of the antenna above mean sea level, in meters.
+  double altitude_meters = 3;
+
+  // The yaw (i.e., left-right rotation), in degrees.
+  // https://en.wikipedia.org/wiki/Euler_angles#Tait-Bryan_angles
+  double yaw_degrees = 4;
```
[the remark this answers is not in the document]
anuraaga: Ugh yeah forgot - thanks for catching

PR #82, opened 2018-11-12
API for add/get TLE

gavinreaney:
I originally had satellite ID on the 'addTle' request but removed it since it is redundant. However, it might be more consistent to keep it there.

### DIFF
--- a/api/src/main/proto/stellarstation/api/v1/stellarstation.proto
+++ b/api/src/main/proto/stellarstation/api/v1/stellarstation.proto
@@ -114,10 +114,10 @@ service StellarStationService {
   // The pass must be scheduled at least one hour in advance of `aos_time` or it will be rejected.
   rpc ReservePass (ReservePassRequest) returns (ReservePassResponse);
 
-  // Add TLE for a satellite.
+  // Set TLE for a satellite.
   //
   // If the TLE is accepted, it will be applied when executing upcoming plans, and for pass
-  // calculation..
+  // calculation.
   //
   // An 'INVALID_ARGUMENT' error will be returned in the following conditions:
   // - the TLE cannot be parsed
@@ -126,7 +126,7 @@ service StellarStationService {
   //
   // If the satellite is not found or the client is not authorized for it, the request will return
   // a `NOT_FOUND` error.
-  rpc AddTle (AddTleRequest) returns (AddTleResponse);
+  rpc SetTle (SetTleRequest) returns (SetTleResponse);
 
   // Gets the current TLE for a satellite.
   //
@@ -471,8 +471,8 @@ message Plan {
   repeated TelemetryMetadata telmetry_metadata = 13;
 }
 
-// Request for the 'AddTle' method.
-message AddTleRequest {
+// Request for the 'SetTle' method.
+message SetTleRequest {
   // The ID of the satellite to add TLE for. The ID can be found on the StellarStation
   // Console page for the satellite.
   string satellite_id = 1;
@@ -481,8 +481,8 @@ message AddTleRequest {
   orbit.Tle tle = 2;
 }
 
-// Response for the `AddTle` method.
-message AddTleResponse {
+// Response for the `SetTle` method.
+message SetTleResponse {
   // Currently no payload in the response.
 }
 
@@ -495,6 +495,6 @@ message GetTleRequest {
 
 // Response for the 'GetTle' method.
 message GetTleResponse {
-  // The latest TLE for the given satellite.
+  // The TLE for the given satellite.
   orbit.Tle tle = 1;
 }
--- a/api/src/main/proto/stellarstation/api/v1/stellarstation.proto
+++ b/api/src/main/proto/stellarstation/api/v1/stellarstation.proto
@@ -114,7 +114,7 @@ service StellarStationService {
   // The pass must be scheduled at least one hour in advance of `aos_time` or it will be rejected.
   rpc ReservePass (ReservePassRequest) returns (ReservePassResponse);
 
-  // Sets the TLE for a satellite.
+  // Set TLE for a satellite.
   //
   // If the TLE is accepted, it will be applied when executing upcoming plans, and for pass
   // calculation.
--- a/api/src/main/proto/stellarstation/api/v1/stellarstation.proto
+++ b/api/src/main/proto/stellarstation/api/v1/stellarstation.proto
@@ -495,5 +495,6 @@ message GetTleRequest {
 
 // Response for the 'GetTle' method.
 message GetTleResponse {
+  // The latest TLE for the given satellite.
   orbit.Tle tle = 1;
 }
--- a/api/src/main/proto/stellarstation/api/v1/stellarstation.proto
+++ b/api/src/main/proto/stellarstation/api/v1/stellarstation.proto
@@ -17,6 +17,7 @@
 syntax = "proto3";
 
 import "google/protobuf/timestamp.proto";
+import "stellarstation/api/v1/orbit/orbit.proto";
 
 package stellarstation.api.v1;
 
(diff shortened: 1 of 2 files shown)


api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -449,3 +469,33 @@ message Plan {
   // successfully and data processing is complete.
   repeated TelemetryMetadata telmetry_metadata = 13;
 }
+
+// Request for the 'AddTle' method.
+message AddTleRequest {
+  // The first line of the TLE. Not a title line.
```
anuraaga: Yeah let's keep `satellite_id` in here for consistency, clarity, and we can use it as a double-check vs the norad ID. It looks like your comment on `AddTle` is already ready for it too :)

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -449,3 +469,37 @@ message Plan {
...
 }
+
+// Request for the 'AddTle' method.
+message AddTleRequest {
+  // The ID of the satellite to add TLE for. The ID can be found on the StellarStation
+  // Console page for the satellite.
+  string satellite_id = 1;
+
+  // The first line of the TLE. Not a title line.
+  string tle_line_1 = 2;
```
anuraaga: Let's go ahead and add an orbit package with a `Tle` message - wish we did it before adding such a message to `groundstation` but probably still a good idea to do this now.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -449,3 +470,30 @@ message Plan {
...
+// Request for the 'GetTle' method.
+message GetTleRequest {
+  // The ID of the satellite for which to fetch TLE. The ID can be found on the StellarStation
+  // Console page for the satellite.
+  string satellite_id = 1;
+}
+
+// Response for the 'GetTle' method.
+message GetTleResponse {
+  orbit.Tle tle = 1;
```
anuraaga: `// The latest TLE for the given satellite.`

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -112,6 +113,26 @@ service StellarStationService {
   // `ListUpcomingAvailablePasses` should be called again to get a fresh list of reservable passes.
   // The pass must be scheduled at least one hour in advance of `aos_time` or it will be rejected.
   rpc ReservePass (ReservePassRequest) returns (ReservePassResponse);
+
+  // Add TLE for a satellite.
+  //
+  // If the TLE is accepted, it will be applied when executing upcoming plans, and for pass
+  // calculation..
```
hoshir: Extra dot.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -112,6 +113,26 @@ service StellarStationService {
...
+  //
+  // An 'INVALID_ARGUMENT' error will be returned in the following conditions:
+  // - the TLE cannot be parsed
+  // - the TLE is older than the current TLE
+  // - the TLE norad ID does not correspond to the requested satellite.
+  //
+  // If the satellite is not found or the client is not authorized for it, the request will return
+  // a `NOT_FOUND` error.
+  rpc AddTle (AddTleRequest) returns (AddTleResponse);
```
hoshir: I feel UpdateTle sounds better which gives an impression that TLE will be updated immediately for the satellite while AddTle sounds it is stored at somewhere in StellarStation, and it is not very clear when it is applied. Wha do you think?

anuraaga: The problem with `UpdateTle` becomes when it sounds like the first TLE cannot come from the method. Can't come up with any better alternative to `AddTle`.

hoshir: How about provideTle? I don't think "Provide" is used very often as a verb in APIs, I heard the word when talking with external people and Ops team.

gavinreaney: In software APIs provide would tend to imply that the method returns (provides) something rather than having something given to it.

How about 'SetTle' or 'SetCurrentTle' or something like that?

The API only allows setting the most recent TLE so I guess these might be OK? (Internally we aren't actually replacing the TLE at all, but that's an implementation detail I guess).

anuraaga: `SetTle` seems good

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
kenichi-fukushima: Please be more specific about definition of `latest`. latest epoch?

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
hoshir: SetTle seems good to me as well.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
gavinreaney: We could also drop 'latest' entirely (especially since the 'add' API is becoming 'set').
// The TLE for the given satellite.

What is being returned here is the 'active' TLE (which happens to be the one with the most recent epoch that stellarstation knows about).

gavinreaney: I dropped latest. The rpc comment mentions that the tle is the 'current' tle which I believe would be understood to mean 'the TLE used by stellarstation when computing the orbit for this satellite'.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -112,6 +113,26 @@ service StellarStationService {
   // `ListUpcomingAvailablePasses` should be called again to get a fresh list of reservable passes.
   // The pass must be scheduled at least one hour in advance of `aos_time` or it will be rejected.
   rpc ReservePass (ReservePassRequest) returns (ReservePassResponse);
+
+  // Sets the TLE for a satellite.
```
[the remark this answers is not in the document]
anuraaga: We decided that the latest TLE epoch is always what we use. If we get a newer TLE epoch from another source it will be used by default unless an even newer epoch is set manually again.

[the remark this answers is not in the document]
anuraaga: It's only to provide fresh TLE - the assumption we decided on is the latest epoch of TLE is always the best regardless of source.

It does bring the good point that we should document the fact that the latest TLE is always used for computations and the provided TLE may be overwritten by newer TLE from another source. It does make the word Set not so good though, I like Add best again...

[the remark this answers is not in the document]
gavinreaney: Updating the documentation to clarify that latest is always used and reverting back to 'add' is easy enough to do.

Implementing revocation of TLE was what I imagined initially but the 'latest is best' requirement means that it isn't really useful to support such a concept (and it needs a lot of backend work). The frontend will work in a similar way to this API AFAIK.

kenichi-fukushima:
What if a user accidentally adds a TLE with epoch in the distant future? Our ops team will handle removing the TLE?

gavinreaney:
Unless we introduce an upper limit on the epoch (and reject setting such a TLE) we'd need ops intervention. Such an upper limit should apply to all TLE sources I think so we should decide on what is needed internally then come back to update the API comments.

hoshir:
TLE is basically created from an observation, not an estimation. When NORAD or someone observe a satellite, they can determine exact position and velocity of the satellite. This is the reason why pass drift time (i.e. positional difference between actual and estimated) is zero when TLE is updated.
Technically, it is possible to estimate future position based on existing TLE as we do in pass computation, and output it as a TLE by setting future epoch time. However, Space track says it is for satellites having long period and it won't apply to our customer. If we want to allow user to add future TLE by propagation by themselves, we need to reconsider pass drift margin since the TLE may exceeds our expected propagation limit.
For now, I think we should start from TLEs by observation, and allow to accept estimated TLE when we have many requests from customers.

PR #84, opened 2018-11-14
Generate API mock for golang

koncha99:


stubs/golang/build.gradle.kts
```diff
@@ -93,6 +99,72 @@ tasks {
         }
     }
 
+    val copyDepsToMockgenGopath by registering(Copy::class) {
```
anuraaga: Add a comment like `// Because mockgen isn't aware of Go modules, we need to prepare a GOPATH for it like this`

PR #85, opened 2018-11-15
Add monitoring events.

anuraaga:
These events will be streamed during the execution of a pass to allow satellite operators to have real-time metrics for monitoring or debugging.

### DIFF
--- a/api/src/main/proto/stellarstation/api/v1/stellarstation.proto
+++ b/api/src/main/proto/stellarstation/api/v1/stellarstation.proto
@@ -17,6 +17,7 @@
 syntax = "proto3";
 
 import "google/protobuf/timestamp.proto";
+import "stellarstation/api/v1/monitoring/monitoring.proto";
 import "stellarstation/api/v1/orbit/orbit.proto";
 
 package stellarstation.api.v1;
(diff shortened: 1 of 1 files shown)


api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -263,6 +264,29 @@ message Telemetry {
   bytes frame_header = 6;
 }
 
+// A monitoring event that occurred during the execution of the plan. Information about the current
+// configuration of the ground station and state of components is returned to provide information
+// that can help to troubleshoot issues with the plan.
+message PlanMonitoringEvent {
+  // The ID of the plan being monitored.
+  string plan_id = 1;
```
hoshir: Just out of curiosity, why did you choose to add plan_id in PlanMinotoringEvent rather than SteramEvent?

anuraaga: I wanted to allow events that aren't tied to a plan. Another option would be adding plan to `StreamEvent` expecting that it's almost always needed and leaving it unfilled when the event isn't for a plan, but I guess it's a bit clearer for fields to be always filled whenever possible

hoshir: I see, thanks for the explanation.

PR #87, opened 2018-11-19
Copy go.mod and go.sum to generated source

koncha99:


stubs/golang/build.gradle.kts
```diff
@@ -88,8 +88,15 @@ tasks {
...
 
+    val copyGoMod by registering(Copy::class) {
+        into("build/generated/proto/main/github.com/infostellarinc/go-stellarstation")
+        from("go.mod")
+        from("go.sum")
+    }
+
     val generateProto by getting(org.curioswitch.gradle.protobuf.tasks.GenerateProtoTask::class) {
         dependsOn(installProtocGoPlugin, installProtoWrap)
+        finalizedBy(copyGoMod)
```
anuraaga: Ah interesting, this looks like a good idea.

Can you also run `go mod vendor` in the directory? Go 1.11 users will be able to use this mod/sum but previous versions will not work well unless we create a vendor directory too.

.gitignore
```diff
@@ -91,6 +91,7 @@ build/Release
 # Dependency directories
 node_modules/
 jspm_packages/
+vendor/
```
anuraaga: Remove this, we can vendor within the build directory itself.

stubs/golang/build.gradle.kts
```diff
@@ -88,6 +88,10 @@ tasks {
         args("install", "github.com/golang/mock/mockgen")
     }
 
+    val runModVendoring by registering(org.curioswitch.gradle.golang.tasks.GoTask::class) {
+        args("mod", "vendor")
```
anuraaga: Run this after `copyGoMod` and run in the build directory

```kotlin
execCustomizer {
  workingDir = "build/generated/proto/main/github.com/infostellarinc/go-stellarstation"
}
```

stubs/golang/build.gradle.kts
koncha99: I added the task which run ```go mod vendor```. But ```-mod=vendor``` option will be needed when building and testing. I will make changes to support this flag on curioswitch's gradle-golang-plugin.

anuraaga: I think you can use `execCustomizer` to override `GOFLAGS` too

stubs/golang/build.gradle.kts
anuraaga: `dependsOn copyGoMod`

PR #88, opened 2018-11-20
Exclude vendor directory when publishing go stub to github

koncha99:


koncha99:
Sorry, I misunderstood. Vendor has already excluded.

PR #96, opened 2018-12-14
New API fields for plan start/end

gavinreaney:


api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -118,6 +118,18 @@ message Plan {
   // Configuration of the radio device used for uplinking to the satellite. Ground stations will
   // need to configure transmission during the plan to match this device.
   radio.RadioDeviceConfiguration uplink_radio_device = 6;
+
+  // The start time of the plan.
```
hoshir: I think definition of start time should be added in the comment so that GS owner understand difference between AOS and start time.

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -119,16 +119,18 @@ message Plan {
   // need to configure transmission during the plan to match this device.
   radio.RadioDeviceConfiguration uplink_radio_device = 6;
 
-  // The start time of the plan.
+  // The start of the time window reserved for the plan.
   //
-  // This value will never be after aos_time. Unlike aos_time, this value is fixed upon plan
-  // creation.
+  // This timestamp is always before aos_time and should be used by the ground station to prepare
+  // for the upcoming plan. Unlike aos_time, this timestamp does not change after the plan is
```
anuraaga: We use ` to surround names of fields in comments. I don't think it actually does anything in documentation, etc but it's good to mark out field names somehow.

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -118,6 +118,20 @@ message Plan {
   // Configuration of the radio device used for uplinking to the satellite. Ground stations will
   // need to configure transmission during the plan to match this device.
   radio.RadioDeviceConfiguration uplink_radio_device = 6;
+
+  // The start of the time window reserved for the plan.
+  //
+  // This timestamp is always before aos_time and should be used by the ground station to prepare
+  // for the upcoming plan. Unlike aos_time, this timestamp does not change after the plan is
+  // reserved.
+  google.protobuf.Timestamp start_time = 7;
```
kenichi-fukushima: Put these new fields directly above/below aos_time and los_time. I think the new fields should come first because they are the single representative time of a plan now. For the same reason comment for aos_time and los_time should be in relation to start_time and end_time, not the other way around.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -501,6 +501,18 @@ message Plan {
   // Metadata for telemetry received during the pass. Only populated when the pass has completed
   // successfully and data processing is complete.
   repeated TelemetryMetadata telmetry_metadata = 13;
+
+  // The start of the time window reserved for the plan.
+  //
+  // Unlike aos_time, this timestamp does not change after the plan is reserved. aos_time will
+  // never be before this timestamp.
+  google.protobuf.Timestamp start_time = 14;
```
kenichi-fukushima: same as above.

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -119,16 +119,18 @@ message Plan {
   // need to configure transmission during the plan to match this device.
   radio.RadioDeviceConfiguration uplink_radio_device = 6;
 
-  // The start time of the plan.
+  // The start of the time window reserved for the plan.
```
hoshir: The comment should also refer that the time contains GS setup time and tear down time.

PR #100, opened 2018-12-17
Added venv to the .gitignore.

hoshir:


anuraaga:
It's probably ok to add this, but isn't it more common to have virtualenvs outside the repository folder?

hoshir:
Yeah, it is, usually. However, since the default directory of project interpreter in PyCharm is the current project directory, it will help a little those who use the default setting.

PR #102, opened 2018-12-17
Updated the authentication description for Python.

hoshir:
Current description of the authentication for Python is confusing since it writes about connecting to fake printing server. This PR update the procedure to connect to the StellarStation production API server.


hoshir:
Updated the Java code example as well.

PR #104, opened 2018-12-18
Added the initial integration test for ListPlans in GroundStationServ…

hoshir:
…ice.

Added the initial integration test class for ListPlans API in GroundStationService. The integration test is expected to be run in test/python directory.


test/python/groundstation/list_plans_test.py
```diff
@@ -0,0 +1,47 @@
...
+
+from google.protobuf.timestamp_pb2 import Timestamp
+from stellarstation.api.v1.groundstation import groundstation_pb2
+
+from conn.factory import StubFactory
+
+GS_ID = '27'
+
+
+class TestListPlans(unittest.TestCase):
```
anuraaga: Can we use pytest? It has many useful features vs std library.

https://docs.pytest.org/en/latest/

hoshir: Thanks, changed to pytest and added very basic tests as the initial version.

test/python/groundstation/list_plans_test.py
```diff
@@ -0,0 +1,38 @@
...
+class TestListPlans(object):
+    def test_list_plans(self, stub_factory):
+        client = stub_factory.get_gs_service_stub()
+
+        from_time = Timestamp(seconds=int(datetime(2018, 12, 1, 0, 0).timestamp()))
+        to_time = Timestamp(seconds=int(datetime(2018, 12, 31, 0, 0).timestamp()))
+
+        request = groundstation_pb2.ListPlansRequest(
+            ground_station_id=GS_ID,
+            aos_after=from_time,
```
anuraaga: I think the idiomatic way may be to do `request.aos_after.FromDateTime(datetime(...))` - does it work?

anuraaga:
```python
request = groundstation_pb2.ListPlansRequest()
request.ground_station_id = GS_ID
request.aos_after.FromDateTime(...)
request.aos_before.FromDateTime(...)
```

hoshir: Thanks, very good to know.

test/python/conftest.py
```diff
@@ -0,0 +1,24 @@
+# Copyright 2018 Infostellar, Inc.
```
anuraaga: Can you name the folder `integration-tests`?

PR #105, opened 2018-12-18
Add ability to generate python API docs.

anuraaga:
The generator seems to sometimes fail on `oneof`. `oneof` comments aren't so useful usually so just removing them for now, but let me know if I should try to keep them.

I'll follow up with deploying this to a site.

image


build.gradle
```diff
@@ -22,17 +22,8 @@ plugins {
 
 conda {
     'miniconda2-build' {
-        packages.add('cmake')
-        pythonPackages.addAll('conan', 'grpcio-tools', 'twine')
-    }
-}
-
-tools {
```
hoshir: Is it intentional to delete the node section?

anuraaga: Yeah - sorry for unrelated change, but updating curiostack allows us to remove it.

PR #108, opened 2018-12-19
Added more tests.

hoshir:
Added more basic integration tests:
- Ground station API
  - ListUnavailabilityWindows
- StellarsSation API
  - GetTle
  - ListPlans
  - ListUpcomingAvailable

### DIFF
--- a/integration-tests/python/groundstation/list_plans_test.py
+++ b/integration-tests/python/groundstation/list_plans_test.py
@@ -20,15 +20,14 @@ from stellarstation.api.v1.groundstation import groundstation_pb2
 GS_ID = '27'
 
 
-class TestListPlans(object):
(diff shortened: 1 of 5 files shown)


hoshir:
The only reason I used the class is to have same test method name, list_plans, which exists both in stellarstation and groundstation API. The API is the only conflicts now, but I thought it will be convenient to use class not to have future naming conflicts, though I don't think there will be too much cases.

anuraaga:
But it's ok if the name is unique within a module right? I don't think we need to worry about the same method name in different modules like we have now.

Functional style is the more idiomatic one for pytest
https://docs.pytest.org/en/latest/fixture.html#fixture

hoshir:
I see. I missed that name need to be within "module". Updated to functional style.

PR #109, opened 2018-12-21
Add recurring pattern to external API.

gavinreaney:
Some points to consider:

- recurring windows are currently not returned. After this PR, the implementation will be updated to return them.

- the external API does not allow creating recurring windows and I don't think we plan to enable that

- I considered using RepeatingPattern instead of RecurringPattern.

### DIFF
--- a/api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
+++ b/api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
@@ -166,10 +166,10 @@ message UnavailabilityWindow {
     NONE = 0;
 
     // The window recurs daily at the same time.
-    DAILY = 3;
+    DAILY = 1;
 
     // The window recurs weekly on the same day.
-    WEEKLY = 1;
+    WEEKLY = 2;
   }
 
   // Details of the recurrence pattern.
--- a/api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
+++ b/api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
@@ -159,6 +159,23 @@ message UnavailabilityWindow {
 
   // End time of the unavailability window.
   google.protobuf.Timestamp end_time = 3;
+
+  // Indicates whether this window is part of a recurring set.
+  enum RecurringPattern {
+    // This is a one-off window that does not recur.
+    NONE = 0;
+
+    // The window recurs daily at the same time.
+    DAILY = 3;
+
+    // The window recurs weekly on the same day.
+    WEEKLY = 1;
+  }
+
+  // Details of the recurrence pattern.
+  //
+  // Note that 'window_id' is the same for all instances of a recurring window.
+  RecurringPattern recurring_pattern = 4;
 }
 
 // A request for a list of unavailability windows for the specified ground station that


api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -159,6 +159,23 @@ message UnavailabilityWindow {
...
   // End time of the unavailability window.
   google.protobuf.Timestamp end_time = 3;
+
+  // Indicates whether this window is part of a recurring set.
+  enum RecurringPattern {
+    // This is a one-off window that does not recur.
+    NONE = 0;
+
+    // The window recurs daily at the same time.
+    DAILY = 3;
```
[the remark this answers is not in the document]
gavinreaney: Ah, that was a copy and paste from an internal variant. I will order them in a more natural way.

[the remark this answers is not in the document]
kenichi-fukushima: It might be off-topic for this PR but a daily window longer than 24 hours and a weekly window longer than 7 days are possible?

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -159,6 +159,23 @@ message UnavailabilityWindow {
...
+
+    // The window recurs weekly on the same day.
+    WEEKLY = 1;
+  }
+
+  // Details of the recurrence pattern.
+  //
+  // Note that 'window_id' is the same for all instances of a recurring window.
```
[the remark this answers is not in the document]
gavinreaney: Yeah, I guess that's true. The alternatives are to do something special with the ID (bit of a hack) or deprecate 'start_time' and 'end_time' and have a repeating 'range' instead.

Or start and end can be the first occurrence with a repeated field for the recurring events (so that first one could be before the range requested).

hoshir: I just want to clarify the behavior. Will the API return instances of recurring windows?
If so, we will need a max_instance field with default value.

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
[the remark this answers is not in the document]
gavinreaney: Hmm. I guess it is possible. In that case, they will just overlap.

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
gavinreaney: Originally I was going to have one UnavailabilityWindow message per 'instance'. However, Dan pointed out that this causes a change in behavior for scripts that might try to delete windows based on collecting the window IDs (all instances in the recurring set will have the same ID). The alternative is to have some repeated field inside the UnavailabilityWindow message itself.

I'm not sure I understand what you mean by max_instance though. Recurring windows are infinite but this API is restricted by the time range in the request.

hoshir: We might be able to add a repeated message in the future named instance like Google Calendar API but for now, I think it's good to just return one UnavailabilityWindow. 
https://developers.google.com/calendar/recurringevents

About max_instance, never mind about it. I totally missed the fact that the request message has start_time and end_time.

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -159,6 +159,23 @@ message UnavailabilityWindow {
...
+
+  // Indicates whether this window is part of a recurring set.
+  enum RecurringPattern {
+    // This is a one-off window that does not recur.
+    NONE = 0;
+
+    // The window recurs daily at the same time.
+    DAILY = 1;
+
+    // The window recurs weekly on the same day.
```
anuraaga: `and the same time`

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
anuraaga: Having expandable windows have a nested message with the recurring seems like a clean API to me. Such structure is more difficult to express in the CLI, but the API's design is the #1 priority.

PR #113, opened 2019-02-01
Extend radio protocol to include CCSDS-specific framing parameters

reiinakano:
Based on our current satellites and radios, I think I can see this as a valid way to represent a communication protocol. Although we currently have data structures for very generic Viterbi and scrambling messages, I think it would make more sense if we let CCSDS users select CCSDS and give them a smaller, more specific set of variations. Eventually, we can add more framing protocols like Intelsat.

Satellites that use their own thing can select bit stream and set at least the sync word. We can extend this as we see fit.


anuraaga:
Will look at the build failure

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -51,11 +51,123 @@ message CommunicationProtocol {
...
+// It allows for a few common variations satellites use such
+// as setting a custom synchronization word, inverting the
+// convolutional coding polynomials, etc. The sequence of
+// error-correcting codes will be applied as they are defined
+// in the CCSDS standard.
+message CCSDS {
+  // Synchronization word parameters. Although CCSDS defines
+  // a particular synchronization word, we let the user select
+  // a custom one.
+  SynchronizationParams synchronization_word = 1;
```
kenichi-fukushima: `synchronization_params` or `synchronization`. `synchronization_word` is contained in it.

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -51,11 +51,123 @@ message CommunicationProtocol {
...
+
+// CCSDS framing https://public.ccsds.org/Pubs/131x0b3e1.pdf
+// It allows for a few common variations satellites use such
+// as setting a custom synchronization word, inverting the
+// convolutional coding polynomials, etc. The sequence of
+// error-correcting codes will be applied as they are defined
+// in the CCSDS standard.
+message CCSDS {
+  // Synchronization word parameters. Although CCSDS defines
+  // a particular synchronization word, we let the user select
```
kenichi-fukushima: Avoid saying `we` in a public API.

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -51,11 +51,123 @@ message CommunicationProtocol {
...
+  // Scrambling parameters
+  ScramblingParams scrambling_params = 3;
+
+  // CCSDS-specific Reed-Solomon parameters
+  message ReedSolomonParams {
+    // Whether or not Reed-Solomon coding is used
+    bool enabled = 1;
+
+    // BlockLengthType
+    enum BlockLengthType {
```
kenichi-fukushima: This enum is not used.

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -51,11 +51,123 @@ message CommunicationProtocol {
...
+  // Reed-Solomon parameters
+  ReedSolomonParams reed_solomon_params = 4;
 }
 
-// Communication of a stream of bits with no framing defined.
+// Bit stream for satellites that don't conform to a particular
+// standard. Usually, only a synchronization word and expected
+// packet length can be set.
 message BitStream {
+  SynchronizationParams synchronization_word = 1;
```
kenichi-fukushima: `synchronization_params` or `synchronization`

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -51,11 +51,125 @@ message CommunicationProtocol {
...
+    // CCSDS framing settings.
+    CCSDS ccsds = 2;
+
+    // Bit stream framing settings.
+    BitStream bit_stream = 3;
   }
 }
 
-// Communication of a stream of bits with no framing defined.
+// CCSDS framing https://public.ccsds.org/Pubs/131x0b3e1.pdf
```
anuraaga: Add an empty line above and below the link

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -51,11 +51,125 @@ message CommunicationProtocol {
...
+// It allows for a few common variations satellites use such
```
anuraaga: These comments line break at a very low width, we're generally aiming for 80 chars or so I think

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -51,11 +51,125 @@ message CommunicationProtocol {
...
+// as setting a custom synchronization word, inverting the
+// convolutional coding polynomials, etc. The sequence of
+// error-correcting codes will be applied as they are defined
+// in the CCSDS standard.
+message CCSDS {
```
anuraaga: Just to confirm, is CCSDS much more commonly used then space packet protocol? I just learned that CCSDS is an organization, not a protocol. But we should align with what's mostly used in practice.

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -51,11 +51,125 @@ message CommunicationProtocol {
...
+  // select a custom one.
+  SynchronizationParams synchronization_params = 1;
+
+  // CCSDS-specific convolutional coding parameters.
+  // Code rate (r): 1/2
+  // Constraint length (K): 7
+  // Connection vectors:
+  //   G1 = 1111001
+  //   G2 = 1011011
+  message ConvolutionalCodingParams {
```
anuraaga: CCSDSConvolutionalCodingParams

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -51,11 +51,125 @@ message CommunicationProtocol {
...
 
-// Communication of a stream of bits with no framing defined.
+// CCSDS framing https://public.ccsds.org/Pubs/131x0b3e1.pdf
+// It allows for a few common variations satellites use such
+// as setting a custom synchronization word, inverting the
+// convolutional coding polynomials, etc. The sequence of
+// error-correcting codes will be applied as they are defined
+// in the CCSDS standard.
+message CCSDS {
+  // Synchronization word parameters. Although CCSDS defines
```
anuraaga: Which page in the PDF docments this? Want to understand it more to flesh out this comment. For now, I think it's something like

```
// Synchronization word parameters. If not set, synchronization will be enabled using the default CCSDS standard synchronization word.
```

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -51,11 +51,125 @@ message CommunicationProtocol {
...
+  }
+
+  // Convolutional coding parameters
+  ConvolutionalCodingParams convolutional_coding_params = 2;
+
+  // CCSDS-specific scrambling parameters.
+  // Additive scrambler with pseudo-random
+  // sequence generated by the polynomial:
+  // h(x) = x^8 + x^7 + x^5 + x^3 + 1
+  message ScramblingParams {
```
anuraaga: CCSDSScramblingParams

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -51,11 +51,125 @@ message CommunicationProtocol {
...
+    // Whether or not scrambling is applied.
+    bool enabled = 1;
+  }
+
+  // Scrambling parameters
+  ScramblingParams scrambling_params = 3;
+
+  // CCSDS-specific Reed-Solomon parameters
+  message ReedSolomonParams {
```
anuraaga: CCSDSReedSolomonParams

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -51,11 +51,125 @@ message CommunicationProtocol {
   oneof Framing {
     // AX.25 protocol settings.
     AX25 ax25 = 1;
+
+    // CCSDS framing settings.
+    CCSDS ccsds = 2;
+
+    // Bit stream framing settings.
```
anuraaga: Oops - we're using `bitstream` in our wording, not `bit stream`. I notice the message that was already there is `BitStream` but luckily it's not used. Can you go ahead and rename it to `Bitstream` and use the one word version everywhere?

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
anuraaga: Ok, if I'm not mistaken the actual words are on in 9.3.5, so we can mention it

```
// Synchronization word parameters. If not set, synchronization will be enabled using the default CCSDS standard synchronization word as in 9.3.5 (Page 9-3) in the specitication.
```

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
anuraaga: I think it's common, it's not a hard recommendation though. The main reason it came to mind is to match the message name to it's primary comment - CCSDS-specific scrambling parameters are CCSDSScramblingParams.

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
anuraaga: @rtubio Can you confirm that "CCSDS framing" is a common term (taking into account CCSDS is the name of an organization). I guess it's similar to IEEE, where we have protocols like IEEE 802.11x. Is there a number to attach to this one, or is it so common to just refer to it as CCSDS without any other specifier?

vaelen: From my experience so far, "CCSDS" is generally how people refer to this part of the standard, to distinguish it from AX.25, CSP, or Intelsat.  The hardware devices we are working with all call it "CCSDS".

reiinakano: Btw, does CSP define framing and error correction? As far as I can tell, CSP (https://github.com/libcsp/libcsp) is also a network layer protocol that does not define synchronization or channel coding.

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
reiinakano: Can we tell if a protobuf is unset? I assume we look at the `enabled` parameter and if `False`, we consider it unset? To be explicit, I worded it as " If disabled, synchronization will be enabled.." but that kind of sounds confusing...

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -51,11 +51,125 @@ message CommunicationProtocol {
   oneof Framing {
     // AX.25 protocol settings.
     AX25 ax25 = 1;
+
+    // CCSDS framing settings.
+    CCSDS ccsds = 2;
+
+    // Bitstream framing settings.
+    Bitstream bit_stream = 3;
```
anuraaga: `bitstream = 3`

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
vaelen: I guess not.  It says it can be used on top of CCSDS, for example.

reiinakano:
> Yeah we know if a protobuf field is unset.

@anuraaga, just to clarify, we know it's unset by checking if the pointer corresponding to a particular field is NULL, right? This is the behavior we're relying on?

anuraaga:
Yeah in Go it's `nil` in Java there are `hasX` methods.

reiinakano:
Ping

Also @anuraaga do you know why the build is failing?

anuraaga:
Not yet going to look at it today

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
rtubio: The term for those message types at the link layer level should be "Transfer Frame", please check the following standard on page 2-4:

https://public.ccsds.org/Pubs/131x0b3e1.pdf

I also agree with @anuraaga that the term "transfer frame" sounds very generic and we might want to use something more specific. In that sense, something like "CCSDS Transfer Frames" might work as well. 

If these parameters are to be used within a CCSDS context (like, for example, when we are using Cortex or other systems), I think that using "transfer frames" is enough since people will already understand that it refers to the CCSDS link layers.

reiinakano: @anuraaga should I rename this to `CCSDSTransferFrame` or `CCSDSTransferFrameSettings`? This is already in a `oneof` called `Framing` so I don't know whether or not this is too redundant.

anuraaga: I think `CCSDSTransferFrame` is fine, the redundancy with `Framing` isn't such a big deal given people would read and understand the word "transfer frame". If we really wanted to reduce the redundancy, the alternative would be keeping as is but making sure to at least use the word transfer frame in all the comments.

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -51,11 +51,137 @@ message CommunicationProtocol {
   oneof Framing {
     // AX.25 protocol settings.
     AX25 ax25 = 1;
+
+    // CCSDS framing settings.
```
anuraaga: Updated all related comments, `CCSDS transfer frame settings`, for example.

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -51,11 +51,137 @@ message CommunicationProtocol {
...
+  CCSDSReedSolomonParams reed_solomon_params = 4;
+
+  // Frame length in bytes
+  uint32 frame_length_bytes = 5;
+
+  // Line Coding used.
+  LineCoding line_coding = 6;
+}
+
+// Bitstream for satellites that don't conform to a particular
```
anuraaga: This and following message still have very short-width line-breaking.

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -51,11 +51,137 @@ message CommunicationProtocol {
...
+// standard. Usually, only a synchronization word and expected
+// frame length can be set.
+message Bitstream {
+  // Synchronization word parameters. Disabled if unset.
+  SynchronizationParams synchronization_params = 1;
+
+  // Frame length in bytes
```
anuraaga: Period at the end of all line comments.

anuraaga: `Frame length, in bytes.`

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -51,11 +51,137 @@ message CommunicationProtocol {
...
+  uint32 frame_length_bytes = 2;
+
+  // Line Coding used.
+  LineCoding line_coding = 3;
+}
+
+// SynchronizationParams contains parameters for frame
```
anuraaga: We don't use golang style, so just something like `Parameters for frame synchronization.`

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -51,11 +51,137 @@ message CommunicationProtocol {
...
+
+  // Line Coding used.
+  LineCoding line_coding = 6;
+}
+
+// Bitstream for satellites that don't conform to a particular
+// standard. Usually, only a synchronization word and expected
+// frame length can be set.
+message Bitstream {
+  // Synchronization word parameters. Disabled if unset.
```
anuraaga: For 'if unset' type of grammar, we start with it since it's the more important part of the sentence.

`If unset, synchronization is disabled.`

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -51,11 +51,137 @@ message CommunicationProtocol {
...
+  // BlockLengthType
+  enum BlockLengthType {
+    // (255, 223, 16) RS code
+    RS_223_255 = 0;
+
+    // (255, 239, 8) RS code
+    RS_239_255 = 1;
+  }
+
+  BlockLengthType block_length_type = 1;
```
anuraaga: Comment

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -51,11 +51,137 @@ message CommunicationProtocol {
...
+    G2_INVERTED_G1 = 2;
+
+    G2_G1 = 3;
+  }
+
+  PolynomialInversion polynomial_inversion = 1;
+
+  // Enum describing different rates for puncturing
+  enum Puncturing {
+    // Puncturing unused
```
anuraaga: `// No puncturing.`

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -51,11 +51,137 @@ message CommunicationProtocol {
...
+message CCSDSConvolutionalCodingParams {
+  // Polynomial inversion
+  enum PolynomialInversion {
+    // CCSDS with normal phase rotation
+    G1_G2 = 0;
+
+    // CCSDS with inverted phase rotation
+    G1_G2_INVERTED = 1;
+
+    G2_INVERTED_G1 = 2;
```
anuraaga: Comments.

PR #117, opened 2019-02-25
Add a plan state message to the external API.

gavinreaney:
This PR is a starting point for communicating plan state from the ground station. I'd like some suggestions on what failure reasons we can add to the enum (or could do that later).


api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
```diff
@@ -231,6 +231,30 @@ message AntennaState {
   antenna.AntennaPolarization polarization = 3;
 }
 
+// The current state of the plan being executed by the ground station.
+message PlanState {
+  // The status of the plan.
+  enum Status {
+    UNKNOWN = 0;
+    STARTING = 1;
```
kenichi-fukushima: What is `STARTING`? Should there be `COMPLETING` too?

gavinreaney: Ah, 'STARTED' seems better.

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
```diff
@@ -236,7 +236,7 @@ message PlanState {
   // The status of the plan.
   enum Status {
     UNKNOWN = 0;
-    STARTING = 1;
+    STARTED = 1;
     EXECUTING = 2;
     COMPLETE = 3;
```
anuraaga: I guess `COMPLETED` is a little better

https://english.stackexchange.com/questions/15400/complete-or-completed

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
```diff
@@ -231,6 +231,30 @@ message AntennaState {
   antenna.AntennaPolarization polarization = 3;
 }
 
+// The current state of the plan being executed by the ground station.
+message PlanState {
+  // The status of the plan.
+  enum Status {
+    UNKNOWN = 0;
+    STARTED = 1;
```
kenichi-fukushima: So what's the difference between `STARTED` and `EXECUTING`? This message is a state, not an event or something.

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
```diff
@@ -236,21 +236,36 @@ message PlanState {
   // The status of the plan.
   enum Status {
     UNKNOWN = 0;
+
+    // Plan execution has started. This will be sent a short time before AOS when the ground
```
anuraaga: `This will be sent only once a short time...`

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
```diff
@@ -236,21 +236,36 @@ message PlanState {
...
+    // station has started preparing for the upcoming plan.
     STARTED = 1;
+
+    // Plan execution is in progress and has not yet reached the COMPLETE or FAILED state. This
+    // status may be sent multiple times.
```
anuraaga: can consider adding more detail with something like `multiple times and acts as a heartbeat for the ground station.`

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
gavinreaney: I initially had only EXECUTING then I added STARTING which became STARTED. 

It's true that as a state, there isn't a difference. I thought it would be clean to have STARTED as a state which is only sent once and it must be sent as soon as the starpass starts processing.

I thought that EXECUTING could be interpreted as being sent at any time (although we could require that it is sent immediately when processing starts too).

Do you prefer to get rid of STARTED?

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
```diff
@@ -236,21 +236,36 @@ message PlanState {
...
     EXECUTING = 2;
-    COMPLETE = 3;
+
+    // Plan execution is complete. This is sent when the ground station has finished executing
+    // the plan and any associated data processing.
```
anuraaga: Should we just be more explicit and say `the plan and uploading all data.`? Data processing is vague and ambiguous.

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
```diff
@@ -236,21 +236,36 @@ message PlanState {
...
   message FailureDetails {
     enum Cause {
       UNKNOWN = 0;
     }
+
     Cause cause = 1;
   }
 
   oneof Details {
+    // When the plan status is FAILED, this field will be populdated with details about the
```
anuraaga: `The details about the cause of a failure if `status` is `FAILED`.

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
```diff
@@ -236,21 +236,36 @@ message PlanState {
   // The status of the plan.
```
anuraaga: `// A status of a plan` and the field should have `// The status of the plan.`. It's pretty redundant but I think we should be extra-pedantic in our developer API.

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
kenichi-fukushima: I have no opinion on whether there should be `STARTED` or something, but if there is I prefer `PREPARING` or `SETTINGUP`. When I saw these enums for the first time I wonder difference between `STARTING` and `EXECUTING`.

Anyway we should use a standardized vocabulary throughout API comments and all other documentation. How do we explain the concept of plan start/end and aos/los in the plans API?

gavinreaney: There is actually a small disconnect when you consider the existing API.

The messages sent on this monitoring stream are from the ground station point of view, whereas they are being received by the satellite operator. The most obvious case is for 'plan start'. For the ground station, plan start is a few minutes before AOS but we haven't communicated that notion of plan start to the satellite operator before.

There is also an existing status enum in the satellite operator 'list plans' API:

    // The plan is reserved for execution in the future at `aos_time`.
    RESERVED = 0;
    // The plan is currently executing and will end in the future at `los_time`.
    EXECUTING = 1;
    // The plan already finished executing successfully.
    SUCCEEDED = 2;
    // The plan failed to execute correctly.
    FAILED = 3;
    // The plan was canceled. This can be explicitly by calling `CancelPlan` or implicit due to
    // cancellation by the ground station in the plan (e.g., due to bad weather).
    CANCELED = 4;
    // The plan completed execution and telemetry data is being processed.
    PROCESSING = 5;
 
Since that enum was written, a few decisions were made:

- SUCCEEDED should become COMPLETED
- PROCESSING no longer needs to be returned

The comments for this existing enum need to be updated to refer to 'start_time' and 'end_time' instead of 'aos_time' and 'los_time'. Let me take care of that in a new PR.

Overall, I think that changing STARTED to PREPARING may be clearer to differentiate the GS meaning of plan start from the one we already use in the satellite operator API.

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
```diff
@@ -231,6 +231,48 @@ message AntennaState {
   antenna.AntennaPolarization polarization = 3;
 }
 
+// The current state of the plan being executed by the ground station.
+message PlanState {
+  // A description of the plan status.
+  enum Status {
+    UNKNOWN = 0;
+
+    // Plan preparation has started. This will be sent only once a short time before AOS when
```
kenichi-fukushima: Can you avoid specifying how many times a specific state value is sent? It's only monitoring. I want to keep exact timing and exact number of monitoring messages as implementation detail. It is possible that this value is sent twice if sampling rate is high and it is not sent if the rate is low.

gavinreaney: I think it is also good to avoid ambiguity so I wanted to capture likely user expectations.

For example:

- I'd expect the state messages to be sent soon after the state transition occurs
- states such as COMPLETED and FAILED I'd expect to receive only once. PREPARING is probably open to either interpretation but 'will be sent when the ground station starts preparing for plan execution' implies one time.

We may have a bit of a different view on the priority of monitoring. I think it is not the same as general logging, debug messages etc. It could potentially be used as input to other systems (and be passed back to the stream API) so it should be seen as more than 'best effort'.

Having said all that, I do prefer to avoid having too many hard limits specified in the API. I'll update the text to something I think is still reasonable.

gavinreaney:
Summarizing a discussion with Keni.

If the newly added thing is really a 'state', then it should be acceptable for the GS to send it multiple times just like the other states.

If we are actually wanting it to be used like an Event then it isn't really ideal to have it being a member of the GroundStationState proto. The more accurate modelling would be to have a new top level message type like 'GroundStationEvent' with 'PlanEvent' as a member. This change would require more implementation effort since all of the layers that currently act as a pass-through for monitoring messages will need to be updated to also handle the new type.

@hoshir do you have any thoughts on this? I like the idea of modeling this more like an event, but not sure it is worth the extra implementation effort for almost the same end result.

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
```diff
@@ -273,28 +296,9 @@ message PlanState {
...
-// State that can be monitored will vary by the specifications of the ground station that executes
-// the pass. Any information that cannot be monitored for the current ground station will not be
-// filled.
-message GroundStationState {
-
-  // The current state of the transmitter in use during a pass. When debugging issues with data
-  // transmission (e.g., no response from a satellite), it can be good to verify the metrics for
-  // the transmitter are within expected or acceptable values.
-  TransmitterState transmitter = 1;
+// Events generated by the ground station during the operation of a pass.
```
anuraaga: generated by a ground station during the operation of a plan.

api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto
```diff
@@ -241,8 +264,8 @@ message PlanState {
     // for plan execution.
     PREPARING = 1;
 
-    // Plan execution is in progress and has not yet reached the COMPLETED or FAILED state. This
-    // status may be sent multiple times and acts as a heartbeat for the ground station.
+    // Plan execution has started. This status is sent when the AOS time for the plan is
```
anuraaga: when passing the AOS time of the plan.

gavinreaney:
I re-arranged things based on the discussion. When making this latest change I was mainly thinking about whether to add something inside 'StreamEvent' or PlanMonitoringEvent. I chose the latter because I found it hard to come up with a separate top level message that is clearly distinct from PlanMonitoringEvent.

I think we can also assume that events always happen in the context of plans for an API used by satellite operators.

PR #121, opened 2019-03-13
Add an API to open streams from ground stations to serve satellite streams.

anuraaga:
This is just symmetric and reverse of a satellite stream. The naming gets very tricky in this direction though, let me know any thoughts you have.


api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -208,3 +226,71 @@ message DeleteUnavailabilityWindowRequest {
...
+  }
+}
+
+// A request to send telemetry received from a satellite to a user.
+message SendTelemetryRequest {
+  // The telemetry being sent.
+  Telemetry telemetry = 1;
+
+  // The ID of the plan the telemetry is being sent for.
+  string plan_id = 2;
```
gavinreaney: I know this is for to keep symmetry with the satellite stream but I guess it would be OK to reverse these.

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -208,3 +226,71 @@ message DeleteUnavailabilityWindowRequest {
...
+
+// A response for the `OpenGroundStationStream` method. The ground station must process a response
+// immediately when received on the stream.
+message GroundStationStreamResponse {
+  // The ID of the plan this response corresponds to. Unfilled if this response does not correspond
+  // to a plan.
+  string plan_id = 1;
+
+  // A unique ID identifying this response. If filled, any `StreamEvent` generated when handling
+  // this GroundStationStreamResponse` must have this value copied into `StreamEvent` must have this
```
gavinreaney: Delete '`StreamEvent` must have this
  // value copied into '

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -208,3 +226,71 @@ message DeleteUnavailabilityWindowRequest {
...
+  oneof Response {
+    // A response containing commands to send to the satellite.
+    SendSatelliteCommandsResponse send_satellite_commands_response = 3;
+  }
+}
+
+// A response containing a command to transmit to the satellite.
+//
+// Next ID: 1
+message SendSatelliteCommandsResponse {
```
gavinreaney: It feels a bit weird to call this a response when it's kind of a request. I suppose we'd use SendSatelliteCommandsRequest directly if we wanted to keep the same name.

Do you think we are likely to have things inside GroundStationStreamResponse that are not some kind of command?

I'm mostly just musing about the name and can live with it as is.

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -208,3 +226,71 @@ message DeleteUnavailabilityWindowRequest {
...
+  // The command frames to send to the satellite. All commands must be transmitted in sequence
+  // immediately. After all commands have been transmitted, telemetry receive will be immediately.
+  repeated bytes command = 2;
```
gavinreaney: Should we start at ID 1 and remove the next ID comment?

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -182,9 +182,7 @@ message SatelliteStreamRequest {
   repeated Framing accepted_framing = 4;
 }
 
-// A request to send commands to a satellite.
-//
-// Next ID: 1
+
```
gavinreaney: Did you mean to remove the comment? I'm also curious why next ID is 1 and not 3.

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
anuraaga: It's a response on the stream, though a request from the satellite operator. I don't think either is easier to understand, but response seems a bit more consistent.

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -208,3 +226,69 @@ message DeleteUnavailabilityWindowRequest {
...
+
+    // An event that occurred while processing the stream. Satellite operators will often use these
+    // events for debugging and monitoring.
+    //
+    // - `command_sent` must be returned right after completing transmission of
+    //    commands received in a `SendSatelliteCommandsResponse`.
+    // - `plan_monitoring_event.ground_station_configuration` must be returned once at the beginning
+    //   of the execution of a plan and contains information about the actual configuration of the
+    //   ground station at the time.
+    // - `plan_monitoring_event.ground_station_state` must be sent continuously throughout the
```
gavinreaney: "must be sent continually' sounds like it should be sent without any interruption. Maybe 'must be sent regularly'? Not sure if we can give an example such as once per second.

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -208,3 +226,69 @@ message DeleteUnavailabilityWindowRequest {
...
+  oneof Response {
+    // A response containing commands to send to the satellite.
+    SendSatelliteCommandsResponse send_satellite_commands_response = 3;
+  }
+}
+
+// A response containing a command to transmit to the satellite.
+message SendSatelliteCommandsResponse {
+  // The command frames to send to the satellite. All commands must be transmitted in sequence
+  // immediately. After all commands have been transmitted, telemetry receive will be immediately.
```
gavinreaney: I don't think the end of the sentence makes sense: 'telemetry receive will be immediately.'

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -67,6 +68,23 @@ service GroundStationService {
...
+  // The first `GroundStationStreamRequest` sent on the stream is used for configuring the stream.
+  // Unless otherwise specified, all configuration is taken from the first request and configuration
+  // values in subsequent requests will be ignored.
+  //
+  // If the ground station is not found or the API client is not authorized for it, the stream will
+  // be closed with a `NOT_FOUND` error.
+  //
+  // Status: ALPHA This API is under development and may not work correctly or be changed in backwards
+  //         incompatible ways in the future.
+  rpc OpenGroundStationStream (stream GroundStationStreamRequest) returns (stream GroundStationStreamResponse);
```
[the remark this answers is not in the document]
anuraaga: This is a good point thanks for bringing it up. It does seem reasonable to leave it as an implementation detail of the clients of whether they implement the functions of a ground station as one process, or multiple processes or even computers. So supporting multiple streams seems to make sense and I lean towards allowing it. However, it means ground stations would be more responsible for taking care of semantics - for example, both streams would receive commands and the client needs to make sure it's handled correctly. While it makes the API more complex to use in such a situation, I think it's reasonable, and maybe not so commonly done either.

What do you think about just adding text here that multiple streams can be opened and commands will be received by all of them?

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -208,3 +226,69 @@ message DeleteUnavailabilityWindowRequest {
...
+// A request for the `OpenGroundStationStream` method.
+message GroundStationStreamRequest {
+  // The ID of the ground station that is opening a stream. The ID of a ground station can be found
+  // on the StellarStation Console page for the ground station.
+  string ground_station_id = 1;
+
+  // A request to send on the ground station stream. The request does not need to be filled on the
+  // first request to `OpenGroundStationStream`.
+  oneof Request {
+    SendTelemetryRequest send_telemetry_request = 2;
```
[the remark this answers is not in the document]
anuraaga: It is a request to send telemetry - the send is not so much a part of the RPC method name. But indeed, this has always been very tricky for me too. We have used `From/To` before and it's not so great either. I think removing Request/Response from the oneof messages like you suggested below might be best.

anuraaga:
Thanks for the comments, let me know if this is better

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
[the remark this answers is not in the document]
kenichi-fukushima: Just adding the text seems sufficient to me.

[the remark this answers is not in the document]
anuraaga: Thanks - added the text and also added a field to help in debugging to allow identifying one of the streams.

PR #123, opened 2019-03-14
Add polarization to radio device configuration.

danada:
I was thinking about using common abbreviations like `LHCP` instead of `LEFT_HAND_CIRCULAR` but it might not be immediately obvious to some people.

@reiinakano please let me know if there are any common polarization I've missed.

### DIFF
--- a/api/src/main/proto/stellarstation/api/v1/radio/radio.proto
+++ b/api/src/main/proto/stellarstation/api/v1/radio/radio.proto
@@ -71,10 +71,10 @@ enum Polarization {
   ANY = 0;
 
   // Horizontal polarization.
-  H = 1;
+  HORIZONTAL = 1;
 
-  // Horizontal polarization.
-  V = 2;
+  // Vertical polarization.
+  VERTICAL = 2;
 
   // Left hand circular polarization.
   LHCP = 3;
--- a/api/src/main/proto/stellarstation/api/v1/radio/radio.proto
+++ b/api/src/main/proto/stellarstation/api/v1/radio/radio.proto
@@ -43,6 +43,9 @@ message RadioDeviceConfiguration {
   // The protocol used by this device when doing RF communication. If unset, the device is only
   // demodulating / modulating without applying any higher-level communication protocol.
   CommunicationProtocol protocol = 4;
+
+  // The polarization used by this radio device.
+  Polarization polarization = 5;
 }
 
 // A communication protocol used with a radio device. These must contain all the parameters
@@ -58,6 +61,27 @@ message CommunicationProtocol {
     // Bitstream framing settings.
     Bitstream bitstream = 3;
   }
+}
+
+// The type of polarization the antenna will use.
+enum Polarization {
+  // Any polarization can be used for communications.
+  ANY = 0;
+
+  // Horizontal polarization.
+  HORIZONTAL = 1;
+
+  // Horizontal polarization.
+  VERTICAL = 2;
+
+  // Left hand circular polarization.
+  LEFT_HAND_CIRCULAR = 3;
+
+  // Right hand circular polarization.
+  RIGHT_HAND_CIRCULAR = 4;
+
+  // Simultaneous left and right hand circular polarization.
+  SIMULTANEOUS_LEFT_HAND_RIGHT_HAND_CIRCULAR = 5;
 }
 
 // CCSDS-specific convolutional coding parameters.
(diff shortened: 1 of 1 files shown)


reiinakano:
I actually have no idea what polarization is as it is not relevant to software radio. :) 

@rtubio, @vaelen, and @CanoRoger can answer.

vaelen:
Tsk, tsk, Rei, it was in my antenna class!

Seriously though, RHCP and LHCP are fine I think.

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -60,6 +63,27 @@ message CommunicationProtocol {
...
+  VERTICAL = 2;
+
+  // Left hand circular polarization.
+  LEFT_HAND_CIRCULAR = 3;
+
+  // Right hand circular polarization.
+  RIGHT_HAND_CIRCULAR = 4;
+
+  // Simultaneous left and right hand circular polarization.
+  SIMULTANEOUS_LEFT_HAND_RIGHT_HAND_CIRCULAR = 5;
```
vaelen: I don't think this is a thing...

anuraaga:
We can go with the same style as the modulations, with keeping the names so acronyms as long as the comments have full names and relevant links to documents

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -43,6 +43,9 @@ message RadioDeviceConfiguration {
   // The protocol used by this device when doing RF communication. If unset, the device is only
   // demodulating / modulating without applying any higher-level communication protocol.
   CommunicationProtocol protocol = 4;
+
+  // The polarization used by this radio device.
+  Polarization polarization = 5;
```
hoshir: Which does this mean,  a polarization used in a communication or polarizations supported by the radio?

hoshir: @rtubio 
- Can vertical and horizontal polarization be used together?
- What does "simultaneous LHCP/RHCP" actually mean? 
Is it "LHCP and RHCP" or  "LHCP or RHCP", or a special polarization that combine LHCP and RHCP?

I am asking the question because I would like to know if we should define polarization  as an array, or single value.

danada: `RadioDeviceConfiguration` is a configuration that will be used for actual communication.

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
danada: Thanks, I think I got confused looking at some documentation. Removed!

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
```diff
@@ -60,6 +63,26 @@ message CommunicationProtocol {
...
 
+// The type of polarization the antenna will use.
+//
+// https://en.wikipedia.org/wiki/Antenna_(radio)#Polarization
+enum Polarization {
+  // Any polarization can be used for communications.
+  ANY = 0;
+
+  // Horizontal polarization.
+  H = 1;
```
anuraaga: `V` and `H` are too short to be meaningful names in code that uses it (and imagine if someone ended up static importing them, ugh...). Let's go with HORIZONTAL and VERTICAL for them, not that long.

api/src/main/proto/stellarstation/api/v1/radio/radio.proto
anuraaga: Oops finally remembered why I'm feeeling dejavu

We have a polarization enum

https://github.com/infostellarinc/stellarstation-api/blob/75e7945ee70becfa692cce891b96bd7ad9193d20/api/src/main/proto/stellarstation/api/v1/antenna/antenna.proto#L48

And we return monitoring information in `AntennaState`

https://github.com/infostellarinc/stellarstation-api/blob/50d93d713cdab00892ec980e7f855998c90e954c/api/src/main/proto/stellarstation/api/v1/monitoring/monitoring.proto#L214

My understanding is polarization is a feature of the antenna, not the radio which is why it's there (we don't get the monitoring data for LHCP vs RHCP from a radio device but from an ACU). I guess polarization should be next to TLE in our protos.

danada:
Closing this since polarization is a property of an antenna, not a radio.

PR #127, opened 2019-04-23
Added an integration test for OpenSatelliteStream.

hoshir:


integration-tests/python/satellite/open_stream_test.py
```diff
@@ -0,0 +1,73 @@
...
+            if expected_status < 0:
+                # Set expected status based on the current value.
+                expected_status = 1 - is_safe_mode
+
+                # Send the command to toggle the state.
+                command = [bytes(b"\x01\x01")]
+                request_queue.put(command)
+            else:
+                assert is_safe_mode == expected_status
+                return
```
anuraaga: Just to confirm, does this cause the program to exit cleanly? I was sort of expecting to have to sent a termination event through the queue to the request iterator.

integration-tests/python/satellite/open_stream_test.py
```diff
@@ -0,0 +1,73 @@
...
+    request_iterator = generate_request(request_queue)
+
+    expected_status = -1
+    for response in client.OpenSatelliteStream(request_iterator):
+        if response.HasField("receive_telemetry_response"):
+            telemetry_data = response.receive_telemetry_response.telemetry.data
+            assert len(telemetry_data) > 1
+
+            # The second last byte of the telemetry indicates the current status of the fake satellite used
+            # in the test. The value is either of 0 or 1.
```
gavinreaney: Is that something you should also assert? (That the value is 0 or 1)?

integration-tests/python/satellite/open_stream_test.py
```diff
@@ -0,0 +1,73 @@
...
+            is_safe_mode = int(telemetry_data[-2])
+            if expected_status < 0:
+                # Set expected status based on the current value.
+                expected_status = 1 - is_safe_mode
+
+                # Send the command to toggle the state.
+                command = [bytes(b"\x01\x01")]
+                request_queue.put(command)
+            else:
+                assert is_safe_mode == expected_status
```
gavinreaney: Can you be certain that the command to toggle the state will also complete before the next telemetry is received?

hoshir: Yes, because telemetries come in every 10 seconds and the next telemetry comes after 10 sec after sending the toggle command.

integration-tests/python/satellite/open_stream_test.py
hoshir: Added assert to check the value.

integration-tests/python/satellite/open_stream_test.py
hoshir: If you mean this test code, yes, it finishes. Is this what you mean by the comment?

PR #144, opened 2019-05-20
Extend the ground station API for clients directly connecting to StarPass

kenichi-fukushima:
I intentionally avoided replicating the existing code in StarPass for event notification but designed a new API from the ground up to keep it minimal.


api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -266,6 +267,14 @@ message GroundStationStreamRequest {
     // - `plan_monitoring_event.ground_station_event` must be sent once at each of the transitions
     //   documented in `PlanLifecycleEvent` when executing a plan.
     StreamEvent stream_event = 4;
+
+    // The status of plan execution. The client can tell the server that it cannot continue
+    // executing the plan by sending a FAILED status. If the status is sent, the server aborts
+    // the current plan and broadcast the status to all the connecting clients.
```
vaelen:
```suggestion
    // the current plan and broadcast the status to all the connected clients.
```

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -266,6 +267,14 @@ message GroundStationStreamRequest {
...
+    //
+    // This field is ignored unless the client is connecting to an endpoint at the edge device
```
vaelen:
```suggestion
    // This field is ignored unless the client is connected to an endpoint at the edge device
```

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -293,11 +302,16 @@ message GroundStationStreamResponse {
...
-  // A payload sent on the stream to be processed by the ground station. Currently only commands to
-  // send to a satellite will be returned to the ground station.
+  // A payload sent on the stream to be processed by the ground station.
   oneof Response {
     // Commands to send to the satellite.
     SatelliteCommands satellite_commands = 3;
+
+    // An important life cycle event of the plan that is currently running.
+    //
+    // This field is filled only when the client is connecting to an endpoint at the edge device
```
vaelen:
```suggestion
    // This field is filled only when the client is connected to an endpoint at the edge device
```

hoshir:
LGTM

api/src/main/proto/stellarstation/api/v1/groundstation/plan-execution.proto
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// An event notifying a plan has ended.
+message PlanDidEnd {
+  // The execution status of the plan.
+  PlanExecutionStatus status = 1;
+}
+
+// A state of the ground station's devices.
+message DeviceState {
```
reiinakano: What about everything else about the plan like antenna ID, satellite ID, channelset protocol, TLE, etc. etc. Don't external plugins need to know about this too?

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -266,6 +267,14 @@ message GroundStationStreamRequest {
...
     //   documented in `PlanLifecycleEvent` when executing a plan.
     StreamEvent stream_event = 4;
+
+    // The status of plan execution. The client can tell the server that it cannot continue
+    // executing the plan by sending a FAILED status. If the status is sent, the server aborts
+    // the current plan and broadcast the status to all the connected clients.
+    //
+    // This field is ignored unless the client is connected to an endpoint at the edge device
+    // (StarPass).
+    PlanExecutionStatus plan_status = 5;
```
gavinreaney: Do you think that the existing stream_event isn't suitable for the GS implementation, or you feel it is just not a good API to begin with?

Can you not use stream_event but just ignore the parts you don't need just now?

I don't feel strongly about it but introducing divergence now might mean more work later.

api/src/main/proto/stellarstation/api/v1/groundstation/plan-execution.proto
kenichi-fukushima: Good question.

I first thought it would make sense that at least one of those new messages have full plan data. I decided not to go that path. `DeviceState` is an descendant of `GroundStationStreamResponse`. The root message has plan id but not full plan data, which implies the full data should be retrieved by calling `ListPlans`.

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
kenichi-fukushima: I decided to add a new field because the comment for `stream_event` says `Satellite operators will often use these events for debugging and monitoring.` The new field is not intended for direct consumption by the satellite operator but by StarPass's internal plan execution logic.

api/src/main/proto/stellarstation/api/v1/groundstation/plan-execution.proto
```diff
@@ -0,0 +1,90 @@
...
+    PlanWillBegin begin = 2;
+
+    // The plan has ended. Sent when the plan's execution state is determined.
+    PlanDidEnd end = 3;
+  }
+}
+
+// An event notifying a plan is beginning.
+message PlanWillBegin {
+  // The desired state of the ground station's devices at each point of time between AOS and LOS.
```
reiinakano: What is the granularity of `DeviceState`?

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
kenichi-fukushima: I decided to add a new field. The main reason is to hide it from the satellite operator. `StreamEvent` is in `transport.proto`, which is shared by satellite operator APIs and ground station APIs.

api/src/main/proto/stellarstation/api/v1/groundstation/plan-execution.proto
kenichi-fukushima: One second.

kenichi-fukushima:
PTAL. I made a signification revision.

api/src/main/proto/stellarstation/api/v1/groundstation/plan-execution.proto
```diff
@@ -0,0 +1,99 @@
...
+
+    // The plan has ended. Sent when the plan's final status is determined. The client should tear
+    // down the ground statoin facility by the plan's `end_time'.
+    PlanDidEnd end = 2;
+  }
+}
+
+// An event notifying a plan has begun.
+message PlanWillBegin {
+  // Predicted coordinates of the tracked satellite for every second between AOS and LOS. This
```
reiinakano: Any reason for 1 second? Does this mean that if a client needs more granularity (anything beyond UHF, I think?), they should ignore this field and get the TLE from the plan ID?

kenichi-fukushima: Thanks for the feedback. I chose 1 second only because it matches the current implementation. What is the desired granularity for higher bands? I can make the granularity configurable but it incurs additional implementation cost. Can we have the client do interpolation on their own?

reiinakano: This is beyond my knowledge. @vaelen @CanoRoger

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -265,6 +266,13 @@ message GroundStationStreamRequest {
     // - `plan_monitoring_event.ground_station_event` must be sent once at each of the transitions
     //   documented in `PlanLifecycleEvent` when executing a plan.
     StreamEvent stream_event = 4;
+
+    // An event generated by the client during a plan that might affect the plan's execution. The
+    // event is consumed by the server and will not be directly forwarded to the satellite operator.
+    //
+    // This field is ignored unless the client is connected to an endpoint at the edge device
+    // (StarPass).
```
gavinreaney: For clients that are not connected at the edge device, wouldn't they expect similar functionality?

api/src/main/proto/stellarstation/api/v1/groundstation/plan-execution.proto
```diff
@@ -0,0 +1,99 @@
...
+  double azimuth = 1;
+
+  // The elevation.
+  double elevation = 2;
+}
+
+// An event generated in the client during a plan.
+message PlanClientEvent {
+  oneof Event {
+    // An even to notify the server that the plan should be aborted. When the server receives this
```
gavinreaney: event

api/src/main/proto/stellarstation/api/v1/groundstation/plan-execution.proto
```diff
@@ -0,0 +1,99 @@
...
+  // The time which the coordinates are for.
+  google.protobuf.Timestamp time = 1;
+
+  // The angle of the satellite.
+  Angle angle = 2;
+
+  // The range rate (change rate of distance) of the satellite. You can use this value to calculate
+  // Doppler shifted communication frequencies using the following formulae.
+  //   uplink:
+  //     shifted_freq = original_freq * (1.0 + range_rage / speed_of_light)
```
koncha99: range_rate ?

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
kenichi-fukushima: This is a tough question. It would help to brainstorm about what the server would do when it receives a plan abortion event.
- Stop processes and destroy resources created for plan execution. [Only at the edge device]
- Move the plan to a failed status.
- Logs the failure and the reason.

The first two actions will be taken whether the plan completes successfully or fails prematurely. The event only helps the actions to occur early. I thought it was worth doing but I'm now not very sure if it would be worth the complexity. Maybe we just need logging?

gavinreaney: Let me take a step back.

Inside StreamEvent there is already another message called PlanLifeCycleEvent. Will that still be sent to the server in the new system? If yes, then this new PlanClientEvent can be edge device only.

Related to the above: since you are also adding a message called PlanLifeCycleEvent in this PR it might be confusing since there is already another message with the same name.

kenichi-fukushima: Good catch :)

`monitoring.PlanLifecycleEvent` is generated by StellarStation and consumed by satellite operators. I actually don't know whether we should require the client of the groundstation API to generate it. At least within scope of this PR, it is StarPass to generate them. However I can't really decide if this will change or not.

It is only a matter of naming. I will take more time to look for better names.

kenichi-fukushima:
To solve the naming problem in an unproductive way, I decided to use `PlanExecutionServerMessage` and `PlanExecutionClientMessage`.

PR #146, opened 2019-05-30
Added README for Python integration tests

hoshir:
- Added README.md to describe purpose of the code
- Fixed tests so that it work in our test environment.

### DIFF
--- a/integration-tests/python/README.md
+++ b/integration-tests/python/README.md
@@ -3,12 +3,13 @@
 This is a set of Python code that tests StellarStation API aiming to show how you can write API clients 
 for StellarStation API.
 
-Since these code are intended to server as sample code of the API, tests will not pass in your environment.
-If you want to run and pass tests, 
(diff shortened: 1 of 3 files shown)


integration-tests/python/README.md
```diff
@@ -0,0 +1,33 @@
+# Integration tests by Python
+
+This is a set of Python code that tests StellarStation API aiming to show how you can write API clients 
+for StellarStation API.
+
+Since these code are intended to server as sample code of the API, tests will not pass in your environment.
```
gavinreaney: Since this code is intended to show real world usage of the API, you will need to make some changes if you want the tests to pass in your environment:
- You need to change satellite ID and ground station ID in the code to match the satellites and ground stations that you own.
- If you don't own any satellites or ground stations, you may need to disable the corresponding tests
- open-stream tests depend on command and telemetry messages, as well as server infrastructure, specific to our test environment

PR #148, opened 2019-05-31
Add SatelliteCoordinates to Plan

kenichi-fukushima:
This should be merged after https://github.com/infostellarinc/stellarstation-api/pull/147.

See STELLAR-2651 for internal discussion.


api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -164,6 +164,11 @@ message Plan {
   // Configuration of the radio device used for uplinking to the satellite. Ground stations will
   // need to configure transmission during the plan to match this device.
   radio.RadioDeviceConfiguration uplink_radio_device = 6;
+
+  // Predicted coordinates of the tracked satellite for every second between AOS and LOS. This
+  // information is intended for calculating device states such as rotator angles and radio
+  // frequencies.
+  repeated SatelliteCoordinates satellite_coordinates = 9;
```
koncha99: Can you update "Next ID" comment ?

api/src/main/proto/stellarstation/api/v1/groundstation/groundstation.proto
```diff
@@ -164,6 +164,11 @@ message Plan {
   // Configuration of the radio device used for uplinking to the satellite. Ground stations will
   // need to configure transmission during the plan to match this device.
   radio.RadioDeviceConfiguration uplink_radio_device = 6;
+
+  // Predicted coordinates of the tracked satellite for every second between AOS and LOS. This
```
reiinakano: I want to revisit "every second" decision. Right now in our flowgraphs we actually interpolate to 20 shifts per second. That might be too much information though? Cc @vaelen

gavinreaney: How about putting the time into the proto instead of having the requirement that these will always be one per second.

Also, can you add a comment to say this field is only populated for future plans?

gavinreaney: Should we allow for cases where this data isn't needed? In other words, should the request have a parameter to indicate whether it should be returned?

FYI: internally the API used to serve this response will need updated since it doesn't return predictions.

gavinreaney: Ah, missed the fact that there is another wrapper around Angle so you can disregard my comment about adding time. However, you could remove the field comment that says it will be every second.

kenichi-fukushima: I think, the fewer knobs our APIs have, the better. I'd like to avoid introducing those parameters if possible.

- For the boolean flag to turn on and off population of this field, how about populating it if and only if a plan is in the future and it is returned from the ground station API? It is intended for plan execution so the condition suffice. We can add the flag when we add support of other use cases.
- For the granularity, I want to have the client do interpolation if it needs higher granularity then every second. I assume this is not a problem from the viewpoint of accuracy. In addition to added complexity, pass computation is still expensive and computing 20x points will surely have big impact on latency.

reiinakano: Sounds good

PR #151, opened 2019-05-31
Upgraded API version in the printing-client

hoshir:


gradle.properties
```diff
@@ -17,3 +17,5 @@
 org.curioswitch.curiostack.version = 0.0.183.1
 
 org.gradle.jvmargs=-Xmx2048m
+
+version = 0.3.0
```
gavinreaney: This looks reasonable but how does it interact with curiostack gradle plugin? IIRC, that also sets the version for the cloud build and also snapshots.

hoshir:
I accidentally closed the PR. I am reopening it.

examples/fakeserver/build.gradle
```diff
@@ -27,7 +27,7 @@ sourceCompatibility = '1.8'
 targetCompatibility = '1.8'
 
 dependencies {
-    compile 'com.stellarstation.api:stellarstation-api:0.0.8'
+    compile 'com.stellarstation.api:stellarstation-api:0.3.0'
```
chokoswitch: Just to follow up on your question on Slack, this file doesn't seem very copy-pastable, so it's probably fine to use `compile project(':api')` here if you wanted.

hoshir: Yeah, updated the file.

Issue #122, opened 2019-03-14
Update Python printing-client to use Queue instead of loop inlining command requests


anuraaga:
Currently, our Python example has a loop with a `sleep` that generates requests. This is not very representative of real use cases, so it would be better if we move the request generating to a separate thread and use `Queue` to pass them into the stream, which is probably the standard pattern for Python request streams.

anuraaga:
Also just for reference, I happened to find this library that implements a simpler API for python streaming. I don't think we should switch to it for examples since it's not official but it could be good to look at.

hoshir:
The printing-client for Python is updated to use Queue.

PR #152, opened 2019-05-31
Changed to use Queue.

hoshir:
Updated printing-client in Python to use Queue rather than generator. Using queue allow users to write code easily that send commands after looking at telemetry received from the stream.

Resolving #122


reiinakano:
This was so timely. Was starting to write a Python client and the previous pattern did not fit at all with my application.

PR #154, opened 2019-06-05
Use AdoptOpenJDK binaries which shouldn't go away and fix first downl…

anuraaga:
…oad doesn't work.

Oracle only distributes OpenJDK for one version now, so safer to use AdoptOpenJDK.

Also removes `cd` to finally fix the behavior when it's first run.


anuraaga:
You're going to have to merge :)

PR #156, opened 2019-06-10
Use AdoptOpenJDK 11 since error prone doesn't support Java 12 yet.

anuraaga:


gavinreaney:
Internally we just switched off the error prone feature that was giving us a problem with Java 12. Something like:

allprojects {
    afterEvaluate {
        tasks.withType(JavaCompile) {
            // This check is broken in Java 12.  See https://github.com/google/error-prone/issues/1257
            options.errorprone.errorproneArgs = ['-Xep:Finally:OFF']
        }
    }
}

However, I didn't encounter that problem when building the API repo so didn't add the workaround.

Do you believe there are deeper problems with Java 12 and error prone that require us to use JDK 11.

anuraaga:
Ah - I was thinking it's better to use JDK 11 (using AdoptOpenJDK for having a stable URL0 since you'd get to keep all the error prone checks without losing anything. 

I don't think CurioStack is ever migrating to JDK 12 since [the preview features are still being iterated on](https://www.infoworld.com/article/3340052/jdk-13-the-new-features-coming-to-java-13.html) and it [would break everything](https://github.com/google/google-java-format/issues/340) anyways.

gavinreaney:
I assume errorprone will be fixed since they were making fixes for JDK 12 before.

JDK 12 does seem a bit pointless without the raw string literals : )

Are AdoptOpenJDK committing to supporting JDK 11 with security updates and so on? If that is fuzzy then I still prefer to use the latest JDK. Otherwise I don't have a strong feeling about it.

anuraaga:
Yeah AdoptOpenJDK is committing to security updates for all Java releases. Actually the main reason they exist is because Oracle won't update even LTS releases like OpenJDK (as such, the binary download link itself is gone), so it's up to the community (AdoptOpenJDK, or otherwise Debian, etc) to provide the updates. The OpenJDK source will continue to receive updates for security and patch releases, but AdoptOpenJDK and OS package managers are the ones that will actually release binaries with it.

https://adoptopenjdk.net/support.html?variant=openjdk11&jvmVariant=hotspot#roadmap

anuraaga:
Going to go ahead and close this since it'll happen automatically, with a slightly different diff, if upgrading to the latest curiostack here.

PR #158, opened 2019-06-18
Add reservation token for each compatible channel set of a pass.

danada:


api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -320,6 +322,22 @@ message Pass {
   // The center frequency, in Hz, for uplinking in this pass. 0 if uplink is not available in
   // this pass.
   uint64 uplink_center_frequency_hz = 10;
+
+  // A mapping of channel set to its unique reservation token.
```
gavinreaney: I think this is the first ever mention of channels or channel set in the external API.

As such, we need to give a lot more information about what this is and how it should be used. For example, how does it related to the existing reservation token? What is a channel_set_id and where does it come from? etc.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -320,6 +322,22 @@ message Pass {
...
+  message ChannelSetToken {
+    // An ID of a compatible channel set.
+    string channel_set_id = 1;
+
+    // A name of a compatible channel set.
+    string channel_set_name = 2;
+
+    // A unique token for this pass, specific to this channel set, that can
+    // be used for scheduling it by copying to ReservePassRequest.reservation_token.
```
gavinreaney: Does the comment on ReservePassRequest also need updated?

danada:
I've split channel set to its own message so it can be used in both pass and plan messages. I've also deprecated the original `reservation_token` since it is not obvious which channel set will be used when reserving a plan.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -286,10 +287,25 @@ message ListPlansResponse {
   repeated Plan plan = 1;
 }
 
+// A channel set defines one or both of the uplink and downlink `RadioDeviceConfiguration`s to be
+// used when executing a pass. A channel set is unique to a specific satellite, and a satellite may
+// have one or more channel sets.
```
gavinreaney: We could say a bit more about the use case of multiple channel sets.

For example, a satellite will have multiple channel sets if it has multiple radio devices for downlink. In this case, it is necessary to select the configuration that will be used during plan execution.

(Or something along those lines - trying to capture the use case that required channel sets to be part of the API).

danada: Thanks for the comment! I've added a common example of TT&C and mission channel sets.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -422,6 +456,9 @@ message Plan {
   // Metadata for telemetry received during the pass. Only populated when the pass has completed
   // successfully and data processing is complete.
   repeated TelemetryMetadata telemetry_metadata = 13;
+
+  // The channel set used to reserve this plan.
+  ChannelSet channel_set = 16;
```
gavinreaney: It's a bit of a pain to manage, but we could mark this as 'alpha' and remove the deprecation from the other token until everything is live in prod.

PR #161, opened 2019-06-20
Set ci env var during snapshot publish.

danada:


cloudbuild/snapshot-publish.yaml
```diff
@@ -24,7 +24,9 @@ steps:
     chmod 400 /root/.ssh/id_rsa &&
     echo -e "Host github.com\n\tStrictHostKeyChecking no\n" >> /root/.ssh/config &&
     ./gradlew :api:artifactoryPublish :stubs:golang:gitPublishPush -Pbintray.user=$$BINTRAY_USER -Pbintray.key=$$BINTRAY_KEY --stacktrace --no-daemon
-  env: ['BINTRAY_USER=istellar-admin']
+  env:
+    - BINTRAY_USER=istellar-admin
+    - CI=true
```
chokoswitch: While I don't think you're using any features related to it in this build, for consistency you could also set `CI_MASTER=true` indicating this is a master build.

chokoswitch: Incidentally, noticed `portal` - presumably that will use this variable at some point :)

PR #164, opened 2019-06-21
Added pass and plan tests

hoshir:


hoshir:
Removed code to create ImmutableList from pass/plan list returned by gRPC since those list has been already immutable.

Could you take a look at the PR again?

integration-tests/java/src/main/java/com/stellarstation/api/test/satellite/PlanManager.java
```diff
@@ -0,0 +1,54 @@
...
+  private final StellarStationServiceBlockingStub client;
+
+  @Inject
+  public PlanManager(StellarStationServiceBlockingStub client) {
+    this.client = client;
+  }
+
+  @Nullable
+  public List<Plan> list(
+      String satelliteId, LocalDateTime aosAfter, LocalDateTime aosBefore, ZoneOffset zone) {
```
gavinreaney: Did you consider using Instant here instead of LocalDateTime + Zone?

I don't feel strongly about it since this is just example code (either could be better depending on the situation).

integration-tests/java/src/integration-test/java/com/stellarstation/api/test/satellite/PlanManagerTest.java
```diff
@@ -0,0 +1,73 @@
...
+    assertThat(plans).isNotNull();
+    assertThat(plans.size()).isNotZero();
+
+    Plan plan = plans.get(0);
+
+    // Check AOS and LOS.
+    LocalDateTime aos = TimestampUtilities.toLocalDateTime(plan.getAosTime(), ZONE_OFFSET);
+    LocalDateTime los = TimestampUtilities.toLocalDateTime(plan.getLosTime(), ZONE_OFFSET);
+    LocalDateTime now = LocalDateTime.now(ZONE_OFFSET);
+    assertThat(aos).isBefore(now);
```
gavinreaney: We're assuming that the server won't return passes that are due to start very soon which could cause this to fail (e.g. within the next few seconds) and also that the server and client don't have significant clock skew. Maybe add a comment?

PR #168, opened 2019-06-25
Refactored open-stream test.

hoshir:
This is for sharing ideas to improve open-stream test to be better as example. The current code has a few issues that arrays to store values are defined in outside of the StreamObserver that is not completely thread safe, and that assertions could start before response stream is closed.

The new code ensure those don't happen though the telemetry tests need more time to wait its completion than now.

I updated code by 
- Splitting tests for TT&C and event verification for simplification.
- Starting tests after response stream is completed.
- Encapsulating arrays into the StreamObserver class.
- Changing responsibility of stopping test from the onNext method in the stream observer to test method.

@gavinreaney What do you think of idea?


integration-tests/java/src/integration-test/java/com/stellarstation/api/test/satellite/SatelliteStreamerTest.java
```diff
@@ -55,67 +56,57 @@ void setUp() {
...
+        if (response.hasReceiveTelemetryResponse()) {
+          ByteString data = response.getReceiveTelemetryResponse().getTelemetry().getData();
+          if (data.size() > 2) {
+            // The second last byte of the telemetry indicates the current state of the
+            // fake satellite used in the test. The value is either of 0 or 1.
+            int state = data.byteAt(data.size() - 2);
+            safeModeStates.add(state);
+
+            // Send a command to toggle the state.
+            SendSatelliteCommandsRequest commandsRequest =
```
gavinreaney: I was expecting some kind of condition to stop us toggling forever. Maybe we should only send this toggle command when safeModeStates.size() == 1? I'm assuming that when you toggle, you will get a future 'onNext' which will then cause us to toggle again...and repeats forever.

integration-tests/java/src/integration-test/java/com/stellarstation/api/test/satellite/SatelliteStreamerTest.java
```diff
@@ -124,19 +115,83 @@ public void onCompleted() {}
             .setEnableEvents(true)
             .build());
 
-    ListenableFuture<Boolean> timeoutFuture =
-        Futures.withTimeout(
-            commandTestFuture, 30, TimeUnit.SECONDS, Executors.newSingleThreadScheduledExecutor());
+    TimeUnit.SECONDS.sleep(20);
```
gavinreaney: I'm wondering if we can get rid of this delay with a slightly different approach.

Inside onNext, I assume we can detect when the test has collected all data that we need. Maybe we can call requestObserverFuture.onCompleted directly from onNext?

If we do that, then the main thread just needs to wait for the timeoutFuture without any calls to sleep.

integration-tests/java/src/integration-test/java/com/stellarstation/api/test/satellite/SatelliteStreamerTest.java
```diff
@@ -124,19 +115,83 @@ public void onCompleted() {}
...
+    EventTestStreamObserver responseObserver = new EventTestStreamObserver();
+    StreamObserver<SatelliteStreamRequest> requestObserver = streamer.openStream(responseObserver);
+
+    requestObserver.onNext(
+        SatelliteStreamRequest.newBuilder()
+            .setSatelliteId(SATELLITE_ID)
+            .setEnableEvents(true)
+            .build());
 
+    TimeUnit.SECONDS.sleep(10);
```
gavinreaney: Same as the comment above. If we call onCompleted inside onNext after the first status arrives, then I think we can get rid of this sleep.

integration-tests/java/src/integration-test/java/com/stellarstation/api/test/satellite/SatelliteStreamerTest.java
```diff
@@ -124,19 +115,83 @@ public void onCompleted() {}
...
+  @Test
+  void events() throws Exception {
+    final SettableFuture<Boolean> testCompletionFuture = SettableFuture.create();
+
+    class EventTestStreamObserver implements StreamObserver<SatelliteStreamResponse> {
+      private List<AntennaState> antennaStates = new ArrayList();
+
+      @Override
+      public void onNext(SatelliteStreamResponse response) {
+        if (response.hasStreamEvent()) {
```
gavinreaney: We could consider checking the size of antennaStates here and only store the result of the first onNext?

As mentioned elsewhere, also calling the request observers onCompleted might be cleaner than the artificial delay.

integration-tests/java/src/integration-test/java/com/stellarstation/api/test/satellite/SatelliteStreamerTest.java
```diff
@@ -124,19 +115,83 @@ public void onCompleted() {}
             .setEnableEvents(true)
```
gavinreaney: I guess you can remove setEnableEvents here (it doesn't do anything but isn't intended to be set for this test).

hoshir:
Thanks for the great comments. I was able to get rid of all sleeps and timeouts by calling `onCompleted()` in the ResponseObserver.

integration-tests/java/src/integration-test/java/com/stellarstation/api/test/satellite/SatelliteStreamerTest.java
```diff
@@ -55,88 +52,136 @@ void setUp() {
...
+
+        if (response.hasStreamEvent()) {
+          antennaState =
+              response
+                  .getStreamEvent()
+                  .getPlanMonitoringEvent()
+                  .getGroundStationState()
+                  .getAntenna();
+
+          onCompleted();
```
gavinreaney: Should be the request observers onCompleted I think.

integration-tests/java/src/integration-test/java/com/stellarstation/api/test/satellite/SatelliteStreamerTest.java
```diff
@@ -55,88 +52,136 @@ void setUp() {
...
+    final SettableFuture<Boolean> testCompletionFuture = SettableFuture.create();
+
+    class EventTestStreamObserver implements StreamObserver<SatelliteStreamResponse> {
+      private AntennaState antennaState;
+      private int totalResponseReceived;
+
+      @Override
+      public void onNext(SatelliteStreamResponse response) {
+        totalResponseReceived++;
+        if (totalResponseReceived > 10) {
```
gavinreaney: If we are only waiting for one thing to be received we could just check the following I think:

if (antennaState != null) {
  return;
}

integration-tests/java/src/integration-test/java/com/stellarstation/api/test/satellite/SatelliteStreamerTest.java
```diff
@@ -55,88 +52,136 @@ void setUp() {
...
+    class TelemetryAndCommandTestStreamObserver implements StreamObserver<SatelliteStreamResponse> {
+      private final List<Integer> safeModeStates = new ArrayList();
+      private int telemetryReceived;
+      private int totalResponseReceived;
+
+      @Override
+      public void onNext(SatelliteStreamResponse response) {
+        if (response.hasReceiveTelemetryResponse()) {
+          totalResponseReceived++;
+          if (totalResponseReceived > 30) {
```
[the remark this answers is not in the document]
hoshir: The response receives telemetry and events. If something is broken in the fake satellite and it wouldn't send telemetry, the response will only receive events infinitely.  This code is to handle the case by stopping the code when it receives more than 30 response.

integration-tests/java/src/integration-test/java/com/stellarstation/api/test/satellite/SatelliteStreamerTest.java
hoshir: The most flaky part of the current fake satellite is accepting command and toggling status. I want to have a concrete code that can stop the code even when it is broken.

hoshir:
Added timeout again to handle the case where no event comes.

PR #169, opened 2019-06-26
Add uplink and downlink RadioDeviceConfiguration fields to ChannelSet.

danada:
I'm starting to think we should just add the full `RadioDeviceConfiguration` into the `ChannelSet`. @hoshir do you think it's valuable to expose the modulation / various radio device settings in this API as well, or would just the uplink/downlink frequency be adequate?

### DIFF
--- a/api/src/main/proto/stellarstation/api/v1/stellarstation.proto
+++ b/api/src/main/proto/stellarstation/api/v1/stellarstation.proto
@@ -300,6 +300,18 @@ message ChannelSet {
 
   // The name used to identify the channel set (for example, TT&C UHF).
   string name = 2;
+
+  // The center frequency used for uplink, in Hz. Set to 0 if the channel set is downlink only.
+  //
+  // Status: ALPHA This API is under development and may not work correctly or be changed in backwards
+  //         incompatible ways in the future.
+  uint64 uplink_center_frequency_hz = 3;
+
+  // The center frequency used for downlink, in Hz. Set to 0 if the channel set is uplink only.
+  //
+  // Status: ALPHA This API is under development and may not work correctly or be changed in backwards
+  //         incompatible ways in the future.
+  uint64 downlink_center_frequency_hz = 4;
 }
 
 // A pass during which a satellite can be communicated with from a given ground station.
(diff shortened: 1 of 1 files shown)


api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -300,6 +300,18 @@ message ChannelSet {
...
+  // Status: ALPHA This API is under development and may not work correctly or be changed in backwards
+  //         incompatible ways in the future.
+  uint64 uplink_center_frequency_hz = 3;
+
+  // The center frequency used for downlink, in Hz. Set to 0 if the channel set is uplink only.
+  //
+  // Status: ALPHA This API is under development and may not work correctly or be changed in backwards
+  //         incompatible ways in the future.
+  uint64 downlink_center_frequency_hz = 4;
 }
```
hoshir: You'll also need to deprecated downlink_center_frequency_hz and uplink_center_frequency_hz in Pass and Plan message.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -300,6 +300,18 @@ message ChannelSet {
 
   // The name used to identify the channel set (for example, TT&C UHF).
   string name = 2;
+
+  // The center frequency used for uplink, in Hz. Set to 0 if the channel set is downlink only.
+  //
+  // Status: ALPHA This API is under development and may not work correctly or be changed in backwards
```
hoshir: We can remove ALPHA since it will be officially supported.

hoshir:
Yes, I think we can add RadioDeviceConfiguration instead.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
danada: I marked these fields `ALPHA` because they are not guaranteed to be implemented at the time of merging. If someone were to use a snapshot, for example, these fields might not be set. Until they are guaranteed to be implemented I would like to keep the `ALPHA` status.

Similarly, it may be too premature to deprecate `downlink_center_frequency_hz` and `uplink_center_frequency_hz` since there is not yet a stable alternative.

Depending on when the next release of the API is we may be able to promote these fields to stable and deprecate the frequency fields.

hoshir: Got it. Let's change it then.

PR #176, opened 2019-07-02
Small fixes

hoshir:
This PR makes two small changes:
1) Avoid integrationTest being skipped by Gradle due to UP-TO-DATE check.
2) Suppressed ProtoTimestampGetSecondsGetNano warning which appears when getSeconds isn't found around getNanos  method. In our code, getSeconds are called in Timestamps.toSeconds. So it is safe to suppress the warning.

### DIFF
--- a/integration-tests/java/build.gradle
+++ b/integration-tests/java/build.gradle
@@ -49,6 +49,7 @@ task integrationTest(type: Test) {
 
     testClassesDirs = sourceSets.integrationTest.output.classesDirs
     classpath = sourceSets.integrationTest.runtimeClasspath
+    outputs.upToDateWhen { false }
 }
 
 dependencies {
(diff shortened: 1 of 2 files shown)


integration-tests/java/build.gradle
```diff
@@ -49,6 +49,7 @@ task integrationTest(type: Test) {
 
     testClassesDirs = sourceSets.integrationTest.output.classesDirs
     classpath = sourceSets.integrationTest.runtimeClasspath
+    outputs.upToDateWhen { false }
```
gavinreaney: Another option is:

./gradlew cleanIntegrationTest integrationTest

But perhaps you can take a quick read of this advice to see if there is a recommended way:
https://blog.gradle.org/stop-rerunning-tests

hoshir: The blog post make sense. We should design how often it should be run. Since this is integration tests, it is supposed to be run at least every day. 
Reflecting back to my original intention, I added it to rerun it during my development, which is not appropriate way to use the function. I should have used  cleanIntegrationTask or --rerun-nstasks instead. 
I removed the code.

PR #179, opened 2019-07-24
Updated dependencies to resolve security alert

hoshir:
Resolved vulnerability about lodash in yarn.lock by upgrading dependencies.


danada:
What deps were upgraded? The lockfile gets generated by yarn after installing packages, but I don't think our package.json was changed.

hoshir:
The security alert was for the version of lodash in yarn.lock which was used in some library we depends on. Since the library that uses lodash does not specify exact version of lodash, running yarn command upgrade the version in the yarn.lock file.

PR #181, opened 2019-08-22
Explain that most-recently added TLE is used instead of TLE with the l…

danada:
…atest epoch.

### DIFF
--- a/api/src/main/proto/stellarstation/api/v1/stellarstation.proto
+++ b/api/src/main/proto/stellarstation/api/v1/stellarstation.proto
@@ -117,9 +117,9 @@ service StellarStationService {
 
   // Adds TLE for a satellite.
   //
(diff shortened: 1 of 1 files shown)


api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -117,9 +117,9 @@ service StellarStationService {
 
   // Adds TLE for a satellite.
   //
-  // The TLE will be used for satellite orbit calculation until it is superseded by a TLE with a
-  // more recent epoch. This happens automatically if StellarStation pulls new TLE data from
-  // external sources, or manually through this API etc.
+  // The TLE will be used for satellite orbit calculation until it is superseded by a newly
+  // inserted TLE. This happens automatically if StellarStation pulls new TLE data from external
```
hoshir: It's not strong opinion, but we could say "newly added TLE" since the API is AddTLE,

PR #190, opened 2019-09-30
Add a fake modem example using the ground station API.

vaelen:
This example uses the ground station API to mimic an external modem.
Some changes are needed to the API to support full functionality, so for now this version sends the same data for all plans.


examples/go/fakemodem/client.go
```diff
@@ -0,0 +1,147 @@
...
+package main
+
+import (
+	"context"
+	"crypto/tls"
+	"log"
+	"time"
+
+	"github.com/golang/protobuf/ptypes"
+	api "github.com/infostellarinc/go-stellarstation/api/v1/groundstation"
```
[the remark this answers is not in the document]
vaelen: I used `goimports` to order these, which is the standard tool for it.  Are we using a different tool in the stellarcli repo?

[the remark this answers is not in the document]
hoshir: I uses GoLand but gofmt behaves the same way: when you put a blank line between third party and ours, gofmt sort the imports by blocks.

examples/go/fakemodem/client.go
```diff
@@ -0,0 +1,152 @@
...
+		c.groundstation = groundstation
+
+		c.connect()
+	}
+
+	stopFunction := func() {
+		log.Printf("Disconnecting from ground station %v (%v).\n", groundstation.Name, groundstation.ID)
+	}
+
+	c.runner.Start(startFunction, stopFunction)
```
hoshir: I may miss something, but I think Client.Connect doesn't have to be run with Runner. Just simply writing connection logic in the startFunction and post logging in the stopFunction makes the logic simpler and testable.

examples/go/fakemodem/client.go
```diff
@@ -0,0 +1,152 @@
...
+
+	conn, err := grpc.Dial(c.groundstation.Address,
+		grpc.WithTransportCredentials(tc),
+		grpc.WithPerRPCCredentials(jwtCreds))
+
+	if err != nil {
+		log.Fatalf("Couldn't connect to ground station: %v\n", err)
+	}
+
+	go func() {
```
hoshir: If connect() is executed without runner, we don't need this go routine, making this function testable.

vaelen: We still need a thread safe way to close out the connection when we are done with it.

examples/go/fakemodem/client.go
vaelen: The runner is used to close out the grpc connection when either the program exits or a second call to `Connect()` is made.  If we change the code so that the client can not be re-connected to a second star pass (which seems logical) then the code can be made simpler by maintaining a reference to the connection and using a mutex to set/retrieve it.

examples/go/fakemodem/config.go
```diff
@@ -0,0 +1,96 @@
...
+
+/*****************
+ * Configuration *
+ *****************/
+
+var DefaultPlanUpdateIntervalInMinutes = 5
+
+// ChannelConfig defines the configuration for a specific
+// communications channel.
+type ChannelConfig struct {
```
hoshir: If "Channel" means ChannelSet in StellarStation API, I slightly prefer using the consistent word, ChannelSet here and other places in the PR.

examples/go/fakemodem/modem.go
```diff
@@ -0,0 +1,271 @@
...
+)
+
+/*********
+ * Modem *
+ *********/
+
+// Modem does all of the actual work.
+type Modem struct {
+	config         *Config
+	groundstations map[string]GroundStationConfig      // Indexed by name
```
hoshir: I guess you only want to connect to one graoundstation and one satellite from the code at a time. If it is true, we may simplify the code by removing map.

vaelen:
I decided to do this another way.

PR #191, opened 2019-10-03
Removed some requirements from stream benchmarker

rockets3600:


rockets3600:
Ricardo discovered some problems with these packages. I found that I don't actually need them so just removed them.

PR #192, opened 2019-10-08
Added counter for out of order data in stream benchmarker

rockets3600:


examples/python/stream_benchmarker/stream_benchmark.py
```diff
@@ -219,7 +219,10 @@ def compile_details(metrics, initial_time):
 
     mbps = (total_data_size / (now - initial_time).seconds) * 8 / 1024 / 1024
 
-    most_recent_time_last_byte_received = metrics[-1].time_last_byte_received.ToDatetime()
+    if len(metrics) > 0:
+        most_recent_time_last_byte_received = metrics[-1].time_last_byte_received.ToDatetime()
+    else:
+        most_recent_time_last_byte_received = "N/A"
```
vaelen: To make parsing the data easier, it might be better to leave the value empty in this case.

examples/python/stream_benchmarker/stream_benchmark.py
```diff
@@ -219,7 +219,10 @@ def compile_details(metrics, initial_time):
 
     mbps = (total_data_size / (now - initial_time).seconds) * 8 / 1024 / 1024
 
-    most_recent_time_last_byte_received = metrics[-1].time_last_byte_received.ToDatetime()
+    if len(metrics) > 0:
+        most_recent_time_last_byte_received = metrics[-1].time_last_byte_received.ToDatetime()
```
vaelen: This might not always be true if there are out of order packets.  It may be better to keep track of this during the loop instead.

PR #201, opened 2019-11-05
Benchmark tool rewritten in Go

rockets3600:


examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,261 @@
...
+				metricTime:            time.Now().UTC(),
+			}
+			md.metrics = append(md.metrics, metric)
+			storeMostRecentTimeLastByteReceived = timeLastByteReceived
+			md.mostRecentTimeLastByteReceived = storeMostRecentTimeLastByteReceived
+		case timeOfReporting := <-ticker.C:
+			if gotFirstMessage {
+				md.timeOfReporting = timeOfReporting
+				outputDetails := md.compileDetails()
+				sessionSummary.totalBytes += outputDetails.totalDataSize
```
vaelen: This can be made into a method on the session summary type that takes an output details type as input.

examples/go/stream_benchmarker/stream.go
```diff
@@ -0,0 +1,208 @@
...
+	ss.conn.Close()
+
+	<-ss.recvLoopClosedChan
+
+	return nil
+}
+
+func (ss *satelliteStream) recvLoop() {
+	// Initialize exponential back off settings.
+	b := backoff.NewExponentialBackOff()
```
hoshir: I don't think we need to reconnect to the stream when it is cut off since some of metric such as bps won't be correct when stream is reconnected.

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,261 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	signalChan := make(chan os.Signal, 1)
+	signal.Notify(signalChan, os.Interrupt)
+	ticker := time.NewTicker(time.Second * time.Duration(*interval))
+	defer ticker.Stop()
+
+	fmt.Println("Listening for messages")
```
vaelen: Use `log.Println` instead.

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,261 @@
...
+	metricsCount                   int64
+	averageDataSize                float64
+	mbps                           float64
+	outOfOrderData                 int64
+}
+
+type sessionSummary struct {
+	firstByteTime                time.Time
+	lastByteTime                 time.Time
+	totalAverageFirstByteLatency time.Duration
```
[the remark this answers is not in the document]
rockets3600: I named it like this because there is some extra logic in the print function:
(totalAverageFirstByteLatency / numIntervals) to get the average for that pass.

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -112,150 +128,250 @@ func (metricsData *metricsData) compileDetails() *outputDetails {
...
-				timeLastByteReceived:  timeLastByteReceived,
-				dataSize:              numBytesInMessage,
-				metricTime:            time.Now().UTC(),
+		case streamResponse := <-streamChannel:
+			message := streamResponse.Telemetry
+			if len(message.Data) != 0 {
+				planID := streamResponse.PlanId
+				if passStarted == false {
+					if planID != "" {
+						log.Println("Receiving messages")
```
vaelen: Add the plan ID here.

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,377 @@
...
+	return sessionSummary
+}
+
+func main() {
+
+	apiKey := flag.String("privateKeyFile", "stellarstation-private-key.json", "api key")
+	endpoint := flag.String("apiEndpoint", "api.stellarstation.com:443", "api endpoint")
+	satelliteID := flag.String("satelliteID", "5", "satellite id")
+	interval := flag.Duration("reportingInterval", 10*time.Second, "reporting interval")
+	assumeEndAfterSeconds := flag.Duration("assumeEndAfter", 10*time.Second, "assume the pass has ended after X duration of no data")
```
vaelen:
```suggestion
	assumeEndAfterDuration := flag.Duration("assumeEndAfter", 10*time.Second, "assume the pass has ended after X duration of no data")
```

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,377 @@
...
+				}
+				md.planID = planID
+			}
+
+		case timeOfReporting := <-reportingTicker.C:
+			if passStarted {
+				md.timeOfReporting = timeOfReporting
+				outputDetails := md.compileDetails()
+				if (*assumeEndAfterSeconds > 0) && (outputDetails.totalDataSize == 0) {
+					secondsSincePassFirstByte := time.Now().UTC().Sub(ps.firstByteTime)
```
vaelen:
```suggestion
					timeSincePassFirstByte := time.Now().UTC().Sub(ps.firstByteTime)
```

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,377 @@
...
+						mostRecentTimeLastByteReceived: md.mostRecentTimeLastByteReceived,
+						initialTime:                    timeOfReporting,
+					}
+				}
+			}
+		case <-signalChan:
+			log.Println("Received interrupt, stopping benchmark")
+			if ps.initialPlanID != "" {
+				passSummaries = append(passSummaries, ps)
+				printPassSummary(output, ps)
```
vaelen: Check for `willPrintPassSummary` here too.

examples/go/stream_benchmarker/conn.go
```diff
@@ -0,0 +1,43 @@
+// Copyright © 2018 Infostellar, Inc.
```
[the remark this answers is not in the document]
vaelen: Still need to update this.

vaelen:
You should probably add a flag for the full summary too.  It can default to true.

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,377 @@
...
+		totalSessionOutOfOrder: totalSessionOutOfOrder,
+		numberOfPlansProcessed: numberOfPlansProcessed,
+	}
+
+	return sessionSummary
+}
+
+func main() {
+
+	apiKey := flag.String("privateKeyFile", "stellarstation-private-key.json", "api key")
```
vaelen:
```suggestion
	apiKey := flag.String("k", "stellarstation-private-key.json", "StellarStation API key file")
```

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,377 @@
...
+	endpoint := flag.String("apiEndpoint", "api.stellarstation.com:443", "api endpoint")
```
vaelen:
```suggestion
	endpoint := flag.String("E", "api.stellarstation.com:443", "API endpoint")
```

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,377 @@
...
+	satelliteID := flag.String("satelliteID", "5", "satellite id")
```
vaelen:
```suggestion
	satelliteID := flag.String("s", "5", "Satellite ID as provided by StellarStation")
```

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,377 @@
...
+	interval := flag.Duration("reportingInterval", 10*time.Second, "reporting interval")
```
vaelen:
```suggestion
	interval := flag.Duration("i", 10*time.Second, "Reporting interval.  (10s, 1m, etc.)  During a pass, an output line will be generated for each reporting interval.")
```

examples/go/stream_benchmarker/benchmark/main.go
vaelen:
```suggestion
	assumeEndAfterDuration := flag.Duration("e", 10*time.Second, "Assume a pass has ended after this much time has passed without receiving any additional data")
```

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,377 @@
...
+	assumeEndAfterSeconds := flag.Duration("assumeEndAfter", 10*time.Second, "assume the pass has ended after X duration of no data")
+	willPrintPassSummary := flag.Bool("passSummary", true, "print a pass summary after every pass")
+	willPrintOverallSummary := flag.Bool("sessionSummary", true, "print an overall summary for all passes")
+	exitAfterPassEnds := flag.Bool("exitAfterPassEnds", false, "exit the program after a pass ends")
```
vaelen:
```suggestion
	exitAfterPassEnds := flag.Bool("x", false, "Exit the program after a pass ends")
```

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,377 @@
...
+}
+
+func main() {
+
+	apiKey := flag.String("privateKeyFile", "stellarstation-private-key.json", "api key")
+	endpoint := flag.String("apiEndpoint", "api.stellarstation.com:443", "api endpoint")
+	satelliteID := flag.String("satelliteID", "5", "satellite id")
+	interval := flag.Duration("reportingInterval", 10*time.Second, "reporting interval")
+	assumeEndAfterSeconds := flag.Duration("assumeEndAfter", 10*time.Second, "assume the pass has ended after X duration of no data")
+	willPrintPassSummary := flag.Bool("passSummary", true, "print a pass summary after every pass")
```
vaelen:
```suggestion
	willNotPrintPassSummary := flag.Bool("P", false, "Do not print a pass summary after each pass")
```

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,377 @@
...
+	willPrintOverallSummary := flag.Bool("sessionSummary", true, "print an overall summary for all passes")
```
vaelen:
```suggestion
	willNotPrintOverallSummary := flag.Bool("S", true, "Do not print an overall summary when the program exits")
```

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,377 @@
...
+	exitAfterPassEnds := flag.Bool("exitAfterPassEnds", false, "exit the program after a pass ends")
+	filename := flag.String("outputFile", "", "specify output file and benchmark will output to it instead of standard out")
```
vaelen:
```suggestion
	filename := flag.String("o", "", "Write report output to a file instead of standard out")
```

rockets3600:
Example output of two passes:
[output.txt](https://github.com/infostellarinc/stellarstation-api/files/3834545/output.txt)

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,412 @@
...
+		ps.firstByteTime.Format(dateTimeFormat),
+		ps.lastByteTime.Format(dateTimeFormat),
+		averageFirstByteLatency,
+		averageLastByteLatency,
+		ps.totalBytes,
+		ps.totalMbps/ps.numIntervals,
+		ps.totalOutOfOrderData)
+}
+
+func (sessionSummary *sessionSummary) printSessionSummary(output io.Writer) {
```
vaelen: Since this is on `sessionSummary`, you can just call it `print(output io.Writer)`.

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,412 @@
...
+		outputDetails.averageFirstByteLatency.Seconds(),
+		outputDetails.averageLastByteLatency.Seconds(),
+		outputDetails.totalDataSize,
+		outputDetails.metricsCount,
+		outputDetails.averageDataSize,
+		outputDetails.dataRateInMbps,
+		outputDetails.outOfOrderData)
+}
+
+func printPassSummary(output io.Writer, ps passSummary) {
```
vaelen: Move this to a method on the `passSummary` struct and rename it `print`.

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,412 @@
...
+		"First byte latency",
+		"Last byte latency",
+		"Total bytes",
+		"Num messages",
+		"Avg bytes",
+		"Mbps",
+		"Out of order\n")
+}
+
+func printPassDetails(output io.Writer, outputDetails outputDetails) {
```
vaelen: Move this to a method on the `outputDetails` struct and call it `print`.

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,412 @@
...
+		averageDataSize:                averageDataSize,
+		dataRateInMbps:                 dataRateInMbps,
+		outOfOrderData:                 outOfOrderData,
+		planID:                         planID,
+	}
+
+	return outputDetails
+}
+
+func printHeader(output io.Writer, planID string) {
```
vaelen: Move this to the `outputDetails` struct since it is related to that data.

examples/go/stream_benchmarker/stream.go
vaelen: We decided to leave the re-connection logic in place with the understanding that metrics may be skewed for a single reporting period, but the overall metrics should still be accurate.

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,261 @@
...
+			fmt.Println("\nReceived interrupt, stopping benchmark")
+			done = true
+		}
+	}
+	if *printSessionSummary {
+		if sessionSummary.numIntervals > 0 {
+			printSummary(&sessionSummary)
+		}
+	}
+	ss.Close()
```
vaelen: Move this to a defer right after you check for the error from creating this object.  That way it gets called even if there is a panic.

rockets3600: I think you might be looking at a previous commit, because there is a defer ss.Close() right after checking for the error from creating SatelliteStream object.

examples/go/stream_benchmarker/stream.go
```diff
@@ -0,0 +1,208 @@
...
+func OpenSatelliteStream(o *SatelliteStreamOptions, recvChan chan<- *stellarstation.Telemetry) (SatelliteStream, error) {
+	s := &satelliteStream{
+		acceptedFraming: o.AcceptedFraming,
+		satelliteId:     o.SatelliteID,
+		apiKey:          o.APIKey,
+		endpoint:        o.Endpoint,
+
+		streamId:           "",
+		recvChan:           recvChan,
+		state:              OPEN,
```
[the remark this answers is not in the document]
rockets3600:
```// Close closes the stream.
func (ss *satelliteStream) Close() error {
	atomic.StoreUint32(&ss.state, CLOSED)

	ss.stream.CloseSend()
	ss.conn.Close()

	<-ss.recvLoopClosedChan

	return nil
}

I don't need the state in the Close function? Admittedly I do not know how it is used when closing the stream.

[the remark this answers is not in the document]
Comment (vaelen) [reply examples/go/stream_benchmarker/stream.go]: I think he meant that if you got rid of the reconnect logic you wouldn't need to keep the state here.  At the moment it is still required.

File: examples/go/stream_benchmarker/benchmark/main.go [hunk 24]
```diff
@@ -0,0 +1,417 @@
...
+	timeOfReporting                time.Time
+	planID                         string
+}
+
+func (metricsData *metricsData) compileDetails() outputDetails {
+	now := time.Now().UTC()
+
+	var averageFirstByteLatency time.Duration
+	var averageLastByteLatency time.Duration
+	var averageDataSize float64 = 0
```
hoshir: Initialization by 0 isn't necessary in Go.

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+
+	satelliteStreamOptions := &benchmark.SatelliteStreamOptions{
+		AcceptedFraming: []stellarstation.Framing{stellarstation.Framing_AX25, stellarstation.Framing_BITSTREAM},
+		SatelliteID:     satelliteID,
+		APIKey:          apiKey,
+		Endpoint:        endpoint,
+	}
+
+	var streamChannel = make(chan *stellarstation.ReceiveTelemetryResponse)
```
hoshir: `streamChannel :=`

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,417 @@
...
+	signalChan := make(chan os.Signal, 1)
+	signal.Notify(signalChan, os.Interrupt)
+	reportingTicker := time.NewTicker(interval)
+	defer reportingTicker.Stop()
+
+	log.Println("Listening for messages")
+	var ps passSummary
+	var md metricsData
+	var passSummaries []passSummary
+	var passStarted = false
```
hoshir: `passStarted := false`
`done` and `printedHeader` can be rewritten with `:=`

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,417 @@
...
+				md.timeOfReporting = timeOfReporting
+				outputDetails := md.compileDetails()
+				if !printedHeader {
+					outputDetails.printHeader(output)
+					printedHeader = true
+				}
+				if (assumeEndAfterDuration > 0) && (outputDetails.totalDataSize == 0) {
+					timeSincePassFirstByte := time.Now().UTC().Sub(ps.firstByteTime)
+					if timeSincePassFirstByte > assumeEndAfterDuration {
+						go func() { passEndedChan <- true }()
```
hoshir: If you want to fire a signal to tell the pass is ended, `close(passEndedChan)` is more natural way to do it.

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,417 @@
...
+					ps.totalBytes += outputDetails.totalDataSize
+					ps.totalMbps += outputDetails.dataRateInMbps
+					ps.totalAverageFirstByteLatency += outputDetails.averageFirstByteLatency
+					ps.totalAverageLastByteLatency += outputDetails.averageLastByteLatency
+					ps.numIntervals++
+					ps.lastByteTime = md.mostRecentTimeLastByteReceived
+					ps.totalOutOfOrderData += outputDetails.outOfOrderData
+				}
+
+				if !done && outputDetails.planID != "" {
```
hoshir: I feel there wound't be a case where done is true. If so, you can omit `!done &&`

examples/go/stream_benchmarker/benchmark/main.go
```diff
@@ -0,0 +1,417 @@
...
+				}
+			}
+		case <-signalChan:
+			log.Println("Received interrupt, stopping benchmark")
+			if !willNotPrintPassSummary && ps.initialPlanID != "" {
+				passSummaries = append(passSummaries, ps)
+				ps.print(output)
+			}
+
+			done = true
```
hoshir: I also suggest to use channel type for `done` and close it when necessary since it is a typical structure in Go and easier to read. The structure will be
```
done := make(chan struct{})
for {
  select {
  case streamResponse := <-streamChannel:
  case timeOfReporting := <-reportingTicker.C:
  case <-signalChan:
  case <-passEndedChan:
    ...
    if exitAfterPassEnds {
      close(done) 
    }
  case <-done:
    break // or write a finalization code
  }
}

examples/go/stream_benchmarker/stream.go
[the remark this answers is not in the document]
hoshir: It looks like this is required for graceful shutdown which ensures the receive logic won't be closed until received data are handled. So, we can keep it.

examples/go/stream_benchmarker/benchmark/main.go
vaelen: When this is left running for multiple passes,  we can't close the channel.

vaelen:
Please run. `golint` and `gofmt -s` on the code. That should fine most of the issues Ryuichi mentioned.

examples/go/stream_benchmarker/benchmark/main.go
hoshir: I didn't recognize the use case, thanks for telling it. 
In that case how about moving the code in `case <-passEndedChan:` here? I guess passEndedChan doesn't have to be a channel since only this line sends true to the channel and it won't happen in parallel. Code to handle passEndedChan can be executed sequentially in this if statement, if I understand correctly, which can reduce an unexpected timing issue caused by using goroutine to send true to the channel.

vaelen: Yeah, good catch.

PR #202, opened 2019-11-18
Added a gradle build file

rockets3600:


examples/go/stream_benchmarker/go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/infostellarinc/stellarstation-api/examples/go/benchmark
+module github.com/infostellarinc/stellarstation-api/examples/go/stream_benchmarker
```
hoshir: This isn't a major comment. I feel like to be consistent whether we prefer `-`  or `_` as a separator of words in the path. Since we have already used `-`, it might be better to use `-` stream_benchmarker.

vaelen: That makes sense.

PR #211, opened 2019-12-23
Rename RZ to DNRZ.

danada:
I haven't been able to find a reliable source for the comment, maybe we shouldn't include the return to zero wikipedia comment. What do you think @vaelen ?


vaelen:
I think what you have is fine.  I found the definition in the manual for a specific hardware modem, so who knows where else it is defined.

PR #214, opened 2020-01-08
Remove deprecated center frequency fields.

danada:
~These fields don't seem to be filled anymore.~

Still being filled, but requesting removal.


danada:
Will merge on 2020-05-01.

PR #215, opened 2020-01-09
Remove deprecated reservation token

danada:


danada:
Will merge on 2020-05-01.

PR #216, opened 2020-01-10
Upgrade to curiostack 0.3.0

danada:
This is still a WIP since the build is failing when trying to generate javadocs on the stubs.

The intentions of the upgrade is to make the developer experience better since there have been many usability improvements since 0.1.0. Since there is no server code we're not really affected by the Armeria changes so I went straight to 0.3.0.


build.gradle
```diff
@@ -15,12 +15,13 @@
  */
 
 plugins {
-    id 'org.ajoberstar.git-publish' version '2.0.0-rc.2' apply false
-    id 'org.curioswitch.gradle-curiostack-plugin'
+    id 'org.ajoberstar.git-publish' version '3.0.0-rc.1' apply false
+    id 'org.curioswitch.gradle-grpc-api-plugin'
```
chokoswitch: You shouldn't need gRPC (unused) or conda (automatically applied) plugins in the root project

stubs/python/build.gradle.kts
```diff
@@ -86,7 +85,7 @@ tasks {
 
     val generateProto = named<org.curioswitch.gradle.protobuf.tasks.GenerateProtoTask>("generateProto")
     generateProto.configure {
-        dependsOn(fillRunProtocScript, prepareSetupPy, ":toolsSetupMiniconda2Build")
```
chokoswitch: You probably need to depend on SetupMinicondaBuild now. Without any dependency the build won't work with a clear cache I think

stubs/docs/build.gradle.kts
```diff
@@ -33,7 +32,7 @@ val GOPATH = extra["gopath"]
 
 protobuf {
     protoc {
-        artifact.set("com.google.protobuf:protoc:3.6.1")
+        artifact.set("com.google.protobuf:protoc:3.11.1")
```
chokoswitch: Does it fail without explicitly setting this?

build.gradle
```diff
@@ -15,12 +15,13 @@
...
+    id 'org.curioswitch.gradle-conda-plugin'
 }
 
 conda {
-    'miniconda2-build' {
+    'miniconda-build' {
```
chokoswitch: Phew looks like this all works with python 3 (good since 2 is permanently dead as of this year).

settings.gradle.kts
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+plugins {
+    id("org.curioswitch.gradle-curiostack-plugin") version "0.3.0"
+}
+
+apply("project.settings.gradle.kts")
```
chokoswitch: Shouldn't need this plugin automatically does it

stubs/python/build.gradle.kts
danada: Ah thanks, was looking for the other instance I removed!

stubs/docs/build.gradle.kts
danada: Thought the was causing javadoc errors at first (since there was that one version of protoc that was broken), but works fine with 3.6.1. Rolled back all instances.

chokoswitch:
LGTM thanks for updating. I realize I was missing a lot in release notes related to the python 3 migration hope it wasn't too much headache.

Also good time to point out (plug ;) this workflow you should be able to copy to get this building on all OSes

https://github.com/curioswitch/curiostack/blob/master/.github/workflows/continuous-build.yml

PR #217, opened 2020-01-17
Add polarization to RadioDeviceConfiguration

koncha99:
It's required to configure devices based on polarization. A ground station that the capability to switch polarization will require this parameter.


danada:
Need to import `stellarstation/api/v1/antenna/antenna.proto`.

PR #223, opened 2020-01-29
Add workflow for building on ubuntu

danada:
Circle CI reporting is weird and I think github actions are cleaner so I'd like to move away from circle CI and just use GH Actions for now. There are some issues in getting other OSes going, but I don't think we lose anything moving away from circle CI.

### DIFF
--- a/.github/workflows/continuous-build.yml
+++ b/.github/workflows/continuous-build.yml
@@ -1,0 +1,61 @@
+# MIT License
+#
+# Copyright (c) 2019 Choko (choko@curioswitch.org)
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-build.yml
```diff
@@ -0,0 +1,61 @@
+# MIT License
+#
+# Copyright (c) 2019 Choko (choko@curioswitch.org)
```
hoshir: Is this file copied from Curiostack? If so, the current copyright statement is right. 
If you wrote this file by yourself, the copyright should be Infostellar.

danada: It's copied from Curiostack in its entirety.

hoshir: That's fine in that case. Thanks.

danada:
Windows and mac builds are failing due to some conan stuff. You can see the broken builds here: https://github.com/danada/stellarstation-api/actions?query=workflow%3A%22Continuous+Build%22

Will continue looking into this.

danada:
Mac build is blocked by #232

danada:
Switched to just ubuntu for now.

PR #224, opened 2020-02-05
Add ground station ID field in Plan message

woobianca:


hoshir:
I think we can add ground_station_id for Pass message in ground station proto, and Plan and Pass for satellite APIs.

PR #233, opened 2020-04-02
Deprecate uplink/downlink radio device in favour of channel set.

danada:
No immediate plan for removal but using `channel_set` will give users some metadata that could be useful (ID, name).


danada:
@koncha99 PTAL 🙏

PR #234, opened 2020-04-06
Address security vulnerability

davidk81:
attempt to address 2 vulnerabilities by running ***yarn upgrade*** (not sure if this is the correct solution)

minimist : https://github.com/infostellarinc/stellarstation-api/network/alert/yarn.lock/minimist/open
kind-of : https://github.com/infostellarinc/stellarstation-api/network/alert/yarn.lock/kind-of/open

vulnerability for *kind-of* may not be resolved because application still depends on ver 3.2.2, 4.0.0, 5.0.0

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -24,11 +24,11 @@ accepts@~1.3.5, accepts@~1.3.7:
     negotiator "0.6.2"
 
 ajv@^6.5.5:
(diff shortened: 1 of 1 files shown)


davidk81:
will not fix because it doesn't actually pose security risks to the usage of the api

PR #235, opened 2020-05-19
Make telemetry repeated

danada:
Motivation here is to improve the performance with high bitrates. On the server side we have to serialize/deserialize thousands of `Telemetry` messages per second and send them one by one which can cause memory issues. 

I propose we set the `Telemetry` message to be repeated to provide (close to) a straight-through path for the telemetry.

⚠️ Although single -> repeated field changes are wire compatible, any clients using old version of the API will only receive the first `Telemetry` message. Another option is to create a new field and deprecate the existing field, but that might have implications on the server since we would have to support both simultaneously (maybe we can do a temporary workaround for high-bitrate satellites).

### DIFF
--- a/integration-tests/java/src/integration-test/java/com/stellarstation/api/test/satellite/SatelliteStreamerTest.java
+++ b/integration-tests/java/src/integration-test/java/com/stellarstation/api/test/satellite/SatelliteStreamerTest.java
@@ -62,7 +62,7 @@ public class SatelliteStreamerTest {
 
     class TelemetryAndCommandTestStreamObserver implements StreamObserver<SatelliteStreamResponse> {
       private boolean isClosing;
-      private final List<Integer> safeModeStates = new ArrayList();
+      private final List<Integer> safeModeStates = new ArrayList<>();
       private int telemetryReceived;
 
       @Override
@@ -72,7 +72,8 @@ public class SatelliteStreamerTest {
         }
 
         if (response.hasReceiveTelemetryResponse()) {
-          ByteString data = response.getReceiveTelemetryResponse().getTelemetry().getData();
+          ByteString data =
+              response.getReceiveTelemetryResponse().getTelemetryList().get(0).getData();
           if (data.size() > 2) {
             // The second last byte of the telemetry indicates the current state of the
             // fake satellite used in the test. The value is either of 0 or 1.
--- a/integration-tests/java/src/integration-test/java/com/stellarstation/api/test/satellite/SatelliteStreamerTest.java
+++ b/integration-tests/java/src/integration-test/java/com/stellarstation/api/test/satellite/SatelliteStreamerTest.java
@@ -27,6 +27,7 @@ import com.stellarstation.api.test.auth.ApiClientModule;
 import com.stellarstation.api.v1.SatelliteStreamRequest;
 import com.stellarstation.api.v1.SatelliteStreamResponse;
 import com.stellarstation.api.v1.SendSatelliteCommandsRequest;
+import com.stellarstation.api.v1.Telemetry;
 import com.stellarstation.api.v1.monitoring.AntennaState;
 import dagger.Component;
 import io.grpc.stub.StreamObserver;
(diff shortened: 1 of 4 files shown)


integration-tests/java/src/integration-test/java/com/stellarstation/api/test/satellite/SatelliteStreamerTest.java
```diff
@@ -72,7 +72,8 @@ public void onNext(SatelliteStreamResponse response) {
         }
 
         if (response.hasReceiveTelemetryResponse()) {
-          ByteString data = response.getReceiveTelemetryResponse().getTelemetry().getData();
+          ByteString data =
```
danada, line 75: Should probably update this similarly to the printing client where all data is combined to a bytestring.

davidk81:
is it feasible / sensible for the client to notify server of version / capabilities ?
this adds complexity during testing, but i think ensuring backwards compatibility seems important in this case

danada:
@davidk81 this is reasonable. It does increase test complexity so let me consider how to accommodate compatibility between API versions. Ultimately we wouldn't make backwards-incompatible changes but it looks like this change (or something similar) may be necessary to accomplish our product's goals.

Let's talk internally!

danada:
Decided that handling this new feature (for now) on the server is the easiest solution. All future users will only have support for multiple messages. We should perform an API release as soon as possible and update [stellarcli](https://github.com/infostellarinc/stellarcli) to handle the repeated `Telemetry` messages.

PTAL @davidk81 @woobianca @koncha99 🙏

davidk81:
do we have a good way to verify this doesn't break compatibility?

danada:
@davidk81 I think it might be possible to add a test that proves the same message can be serialized and deserialized on different versions of the API but I'm wondering how to manage deps in this case.

Locally it's easy to verify by duplicating `ReceiveTelemetryResponse`, leaving one `Telemetry` field singular and making the other one `repeated`, then serializing the `ReceiveTelemetryResponse ` message and deserializing using the `ReceiveTelemetryResponse` with the `repeated` field.

danada:
Here's a little test scenario that is passing. One thing to notice is that the **last** element added to the repeated message is what ends up in the non-repeated message. We wouldn't expect to ever hit this situation, however. The most likely case is that clients on older APIs will continue to receive single messages in a repeated field which works just fine (see final assertion).

```protobuf
message SatelliteStreamResponse {
  string stream_id = 1;

  oneof Response {
    ReceiveTelemetryResponse receive_telemetry_response = 2;
    StreamEvent stream_event = 3;
  }
}

// Non-repeated.
message SatelliteStreamResponse2 {
  string stream_id = 1;

  oneof Response {
    ReceiveTelemetryResponse2 receive_telemetry_response = 2;
    StreamEvent stream_event = 3;
  }
}

message ReceiveTelemetryResponse {
  repeated Telemetry telemetry = 1;
  string plan_id = 2;
}

// Non-repeated.
message ReceiveTelemetryResponse2 {
  Telemetry telemetry = 1;
  string plan_id = 2;
}
```

Java test.
```java
    SatelliteStreamResponse2 singular =
        SatelliteStreamResponse2.newBuilder()
            .setReceiveTelemetryResponse(
                ReceiveTelemetryResponse2.newBuilder()
                    .setTelemetry(
                        Telemetry.newBuilder()
                            .setFraming(Framing.BITSTREAM)
                            .setData(ByteString.copyFromUtf8("baddad"))))
            .build();

    SatelliteStreamResponse repeated =
        SatelliteStreamResponse.newBuilder()
            .setReceiveTelemetryResponse(
                ReceiveTelemetryResponse.newBuilder()
                    .addTelemetry(
                        Telemetry.newBuilder()
                            .setFraming(Framing.BITSTREAM)
                            .setData(ByteString.copyFromUtf8("baddad"))))
            .build();

    SatelliteStreamResponse repeatedMany =
        SatelliteStreamResponse.newBuilder()
            .setReceiveTelemetryResponse(
                ReceiveTelemetryResponse.newBuilder()
                    .addTelemetry(
                        Telemetry.newBuilder()
                            .setFraming(Framing.BITSTREAM)
                            .setData(ByteString.copyFromUtf8("333333")))
                    .addTelemetry(
                        Telemetry.newBuilder()
                            .setFraming(Framing.BITSTREAM)
                            .setData(ByteString.copyFromUtf8("c0ffee")))
                    .addTelemetry(
                        Telemetry.newBuilder()
                            .setFraming(Framing.BITSTREAM)
                            .setData(ByteString.copyFromUtf8("baddad"))))
            .build();

    // Repeated to non-repeated.
    assertThat(SatelliteStreamResponse2.parseFrom(repeated.toByteString())).isEqualTo(singular);
    assertThat(SatelliteStreamResponse2.parseFrom(repeatedMany.toByteString())).isEqualTo(singular);

    // Non-repeated to repeated.
    assertThat(SatelliteStreamResponse.parseFrom(singular.toByteString())).isEqualTo(repeated);
```

davidk81:
great, i guess we should also include a test during QA using pre-change CLI version with post-change server to be sure. and hopefully the java stubs behaves.

PR #236, opened 2020-05-26
STELLAR-3547 - add support for stream resume on index

davidk81:
- welcoming better naming for variables

This adds support for stream resume on errors. The idea is if client opens a stream with `SatelliteStreamRequest.resume_telemetry_index > 0`, streamer-server will attempt to resume telemetry transfer at the requested index.

Using this feature requires also `SatelliteStreamRequest .enable_flow_control = true` and client will need to respond each `ReceiveTelemetryResponse` with a corresponding `ReceiveTelemetryAck`. This is necessary because streamer-server cannot reliably determine which telemetries that has been received correctly by client (due to various errors such as client app crashing, server crashing, etc)


api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -195,11 +195,34 @@ message SatelliteStreamRequest {
     // Status: ALPHA This API is under development and may not work correctly or be changed in backwards
     //         incompatible ways in the future.
     GroundStationConfigurationRequest ground_station_configuration_request = 7;
+
+    // A ack message to respond to each ReceiveTelemetryResponse when `enable_flow_control` is true.
```
danada, line 199:
```suggestion
    // An ack message to respond to each ReceiveTelemetryResponse when `enable_flow_control` is true.
```

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -195,11 +195,34 @@ message SatelliteStreamRequest {
...
+  // than zero, server will attempt to resume transfer starting from the specified index. Otherwise, 
+  // server will attempt to resume transfer with first avaiable message. First index of a pass
+  // always starts at one. This feature may not be available for Out-of-order-allowed-satellites.
+  // This feature is only available if enable_flow_control is true.
+  //
+  // Status: ALPHA This API is under development and may not work correctly or be changed in backwards
+  //         incompatible ways in the future.
+  google.protobuf.Int64Value resume_telemetry_index = 8;
+
+  // Whether flow control is enabled on the client side. If set to true, client must respond each
```
danada, line 220:
```suggestion
  // Whether flow control is enabled on the client side. If set to true, client must respond to each
```

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
davidk81: i think we should able to achieve the same level of performance we have right now, as long as we are generous with the 'inflight' count.

i'm thinking this would replace the current "max_inflight" mechanisms because what we have now is an inaccurate count.

davidk81:
@danada maybe you are more familiar with armeria/grpc, but i was unable to find a way to accurately determine what stream messages have been received. i've concluded myself it cannot  be reliably determined using the "onNext(msg)" and "onReady()" mechanisms provided.

koncha99:
As far as I understand, grpc streaming API doesn't provide a mechanism for guaranteed delivery on application layer. It needs to implement ACK protocol by ourselves like this.

davidk81:
> I'm wondering what this would entail for performance, especially on higher-latency connections.

@danada - i've completed an implementation of streamer-server with this api change.

low latency test (CLI : tokyo office) was able to achieve > 250 Mbps with matched upload speed (low out-of-order scenario)
`maxInflightBufferSizeMb: 64`

image

davidk81:
here's a test with high-latency client
image

danada:
Are the breaks in stats at ~15:04 and ~15:07 caused by reconnecting to the stream?

davidk81:
> Are the breaks in stats at ~15:04 and ~15:07 caused by reconnecting to the stream?

no, in this case probably throughput stalled waiting for the correct message to appear (deduced from 0 ack rage and high nack rate). the line is broken because i had a ">0" filter on the graphed metric.

davidk81:
i've simplified the message ack mechanism; now client just needs to respond with the provided `message_ack_id` instead of supplying both `plan_id` and `index`

on the server side, this id will likely be composed of `sat_id:plan_id:index`, but also gives us more flexibility to change in future.

i've also updated the sample client code with enable_flow_control = true.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -195,11 +195,36 @@ message SatelliteStreamRequest {
...
+  // in attempt to resume streaming after a connection or application error. If the specified,
+  // server will attempt to resume transfer assuming the provided value is that of the most recent Telemetry
+  // received & processed by the client.
+  //
+  // This feature is only available if enable_flow_control is true.
+  // This feature may not be available for Out-of-order-allowed-satellites.
+  //
+  // Status: ALPHA This API is under development and may not work correctly or be changed in backwards
+  //         incompatible ways in the future.
+  string resume_stream_message_ack_id = 8;
```
danada, line 220: Is `stream_id` also required to use this feature? On the CLI we send `stream_id` when reconnecting, but API clients need to do this explicitly.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -195,11 +195,36 @@ message SatelliteStreamRequest {
...
+    //         incompatible ways in the future.
+    ReceiveTelemetryAck telemetry_received_ack = 9;
   }
 
   // The `Framing` types to accept, for satellites that have been configured for multiple framings
   // (e.g., IQ + AX25). If empty, all framings produced by the satellite will be returned.
   repeated Framing accepted_framing = 4;
+
+  // Optional. The `Telemetry.message_ack_id` of the most recently received telemetry. This may be specified
+  // in attempt to resume streaming after a connection or application error. If the specified,
```
danada, line 211:
```suggestion
  // in attempt to resume streaming after a connection or application error. If specified,
```

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -195,11 +195,36 @@ message SatelliteStreamRequest {
...
+  // server will attempt to resume transfer assuming the provided value is that of the most recent Telemetry
```
danada, line 212:
```suggestion
  // the server will attempt to resume transfer assuming the provided value is that of the most recent Telemetry
```

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -195,11 +195,36 @@ message SatelliteStreamRequest {
...
+  // received & processed by the client.
```
danada, line 213: What happens if an invalid ID is provided?

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
davidk81: `stream_id` is not required (nor checked). if any errors occur during parsing, server should function as if value is not provided.

danada: Cool!

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
davidk81: added this to comments

```
If not specified (or if provided value is invalid), server will resume streaming from the most recent telemetry message sent by the server which may or may not have been received by the client.
```

danada: Thanks, makes sense.

davidk81:
any more feedback on getting this merged?

I've completed and tested the Streamer-server and CLI changes based on this, and i'm pretty happy with the results. no negative performance impacts with correct flow control windows, and stream was able to resume multiple times during a pass with no data loss.

PR #237, opened 2020-05-27
Update maven developers

danada:
(mostly a check to see if actions are performed properly)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -86,7 +86,7 @@ allprojects {
                             }
                             developers {
                                 developer {
-                                    id = 'dan'
+                                    id = 'danada'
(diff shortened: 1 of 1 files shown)


danada:
Hmm, wondering if we need both pull and PR triggers.

build.gradle
```diff
@@ -86,9 +86,9 @@ allprojects {
                             }
                             developers {
                                 developer {
-                                    id = 'rag'
-                                    name = 'Anuraag Agrawal'
-                                    email = 'rag@istellar.jp'
+                                    id = 'dan'
```
danada, line 89:
```suggestion
                                    id = 'danada'
```

PR #241, opened 2020-06-22
fake api server - add plan id & msg ack id

davidk81:
- Update fake-api-server to add support for TelemetryReceivedAck. Also, plan-id is now sent with Telemetry Data to emulate real server message.
- This also fixes `--stats` feature for stellarcli when connecting to fake-api-server

- note on `- Echos back received commands` 
Should we remove this? I believe this doesn't happen in our actual server.

### DIFF
--- a/examples/fakeserver/src/main/java/com/stellarstation/api/fakeserver/FakeStellarStationService.java
+++ b/examples/fakeserver/src/main/java/com/stellarstation/api/fakeserver/FakeStellarStationService.java
@@ -52,8 +52,8 @@ import org.apache.logging.log4j.Logger;
 class FakeStellarStationService extends StellarStationServiceImplBase {
   private final FakeServerConfig config;
   private static final Logger logger = LogManager.getLogger();
-  final static String SATELLITE_ID = "5";
-  final static String FAKE_PLAN_ID = System.currentTimeMillis() + "";
+  static final String SATELLITE_ID = "5";
+  static final String FAKE_PLAN_ID = System.currentTimeMillis() + "";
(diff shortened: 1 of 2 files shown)


examples/fakeserver/src/main/java/com/stellarstation/api/fakeserver/FakeStellarStationService.java
```diff
@@ -81,11 +87,15 @@
...
       @Override
       public void onNext(SatelliteStreamRequest value) {
-        if (!value.getSatelliteId().equals("5")) {
+        if (!value.getSatelliteId().equals(SATELLITE_ID)) {
           throw new StatusRuntimeException(Status.INVALID_ARGUMENT);
         }
-        for (ByteString payload : value.getSendSatelliteCommandsRequest().getCommandList()) {
-          sendTelemetry(payload, responseObserver);
+        if (value.hasTelemetryReceivedAck()) {
+          logger.info("received TelemetryReceivedAck message: \n" + value);
```
koncha99, line 94: Could this cause a large amount of logs ?

davidk81: yes it does, I figured this is fake-api-server so should be ok (?). the justification for adding logging is to allow clients to see if they are implementing message acks correctly. let me know if you think we should not log this.

davidk81:
@danada i will leave the command echo-ing as is, I guess we can do something about it if it actually causes confusion for users.

PR #251, opened 2020-11-11
Add modulation to receiver and transmitter configuration requests.

danada:


api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -310,17 +312,25 @@ message TransmitterConfigurationRequest {
 
   // Bitrate of the transmitter
   google.protobuf.FloatValue bitrate = 5;
+
+  // Modulation of the transmitter.
+  radio.Modulation modulation = 6;
```
koncha99, line 317: Can this be null if it's not set？I use wrapper values because star-receiver needs to distinguish what values actually has to be changed.

danada: Good point. Enums can't be null, but default value is `DISABLED`. Maybe we can just ignore `DISABLED` modulation when it is set through a GS configuration request (with comments to explain this behaviour).

koncha99 (approved):

PR #264, opened 2021-04-05
Publish stub to Maven Central

koncha99:
Due to the EOL of bintray and jcenter, this will upload both snapshot and release stub to maven central directly.

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -15,9 +15,9 @@
  */
 
 plugins {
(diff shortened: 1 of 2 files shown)


danada:
What's required to publish with this method? It looks like we need sonatype credentials as well as some configuration for signing. 

We can test this locally (publishing snapshots, etc), however integrating this into our automation will need gpg key management.

danada:
Closed in favour of #266

PR #272, opened 2021-05-17
Replace artifactory references with sonatype repo. Remove incorrect i…

### LINKED ISSUE #271: References to jfrog snapshot repository should be replaced with sonatype
We switched from jfrog to sonatype to host our snapshots, readmes (root and stubs) should be updated to reflect this.

danada:
…nformation about publishing to jcenter, etc.

Information is still not complete, but it is good to get rid of the misleading stuff while we're in transition to a new publishing pipeline.

Closes #271


danada:
Build is failing due to bintray usage in curiostack. Will bump the version shortly.

meierw (approved):

PR #276, opened 2021-05-18
Upgrade to curiostack 0.8.0.

danada:
Release notes: https://github.com/curioswitch/curiostack/releases/tag/curiostack-0.8.0


danada:
Some cpp build is failing, will play around with it but may just disable the task for now since we're not publishing cpp stubs at the moment.

api/build.gradle
```diff
@@ -22,9 +22,6 @@ plugins {
 
 archivesBaseName = 'stellarstation-api'
 
-sourceCompatibility = '1.8'
```
chokoswitch, line 25: I wonder if this change is due to Gradle 7 incompatibility. You should move into a `java` block

https://github.com/aws-observability/aws-otel-java-instrumentation/blob/main/build.gradle.kts#L79

Currently it seems like users of this API module will need to be using Java 16

chokoswitch: Added to the release notes that it's best to check for Gradle 7 deprecations before upgrading :)

danada: Thanks for the tip. Had to bump fakeserver up to java 11 due to curio-server-framework being compatible with java 11: https://github.com/curioswitch/curiostack/blob/7660252d7a9164f60d58713a5953786fbc46bdae/common/server/framework/build.gradle#L31

PR #305, opened 2022-10-13
Removed cpp stubs generation

rockets3600:
This fixes the build issue.
These stubs were being built, but not published anywhere.  I think safe to remove.


adamlb94 (approved):
Nice! (This builds for me locally 👍)

PR #320, opened 2023-03-19
[STS-958] Add aos_before, aos_after, antenna_ids to ListUpcomingAvailablePassesRequest

adamlb94:


RyanProffitt:
As talked about earlier, let's use the field names

string satellite_id = 1;
google.protobuf.Timestamp aos_after = 2;
google.protobuf.Timestamp aos_before = 3;
repeated string antenna_ids = 4;

RyanProffitt (approved):
Looks good to me.

api/src/main/proto/stellarstation/api/v1/stellarstation.proto
```diff
@@ -417,6 +417,21 @@ message ListUpcomingAvailablePassesRequest {
   // The ID of the satellite to list passes for. The ID of a satellite can be found on the
   // StellarStation Console page for the satellite.
   string satellite_id = 1;
+
+  // The start time of the range of plans to list (inclusive). Only passes with an Acquisition of
+  // Signal (AOS) at or after this time will be returned. It is an error for the duration between
+  // `aos_after` and `aos_before` to be longer than 31 days.
+  google.protobuf.Timestamp aos_after = 2;
```
rockets3600, line 424: Are these optional too?
We should support the previous method of only supplying the satelliteId as well.

rockets3600 (approved):

PR #324, opened 2023-12-21
Added in Mik's changes and an update to the README as well.

RyanProffitt:
Making some change to how the example code handles which API to hit.


Vui-Chee (commented):
What if you forget to specify the port in the URL? Do you really want to specify the port or can have that separately?

Vui-Chee (approved):

PR #330, opened 2025-05-01
update to go 1.23

JamieCrisman:
bump up to go 1.23

- grpc-go only supports last 2 versions. 
- move to current protobuf version
- remove gomock (was deprecated)
- remove protowrap (deprecated)
- remove vendored dependencies


stubs/golang/build.gradle.kts
```diff
@@ -80,19 +77,13 @@ tasks {
...
-        args("install", "github.com/square/goprotowrap/cmd/protowrap")
+        args("install", "google.golang.org/protobuf/cmd/protoc-gen-go@latest")
     }
 
-    val installMockGen by registering(org.curioswitch.gradle.golang.tasks.GoTask::class) {
+    val installProtocGoGrpcPlugin by registering(org.curioswitch.gradle.golang.tasks.GoTask::class) {
         dependsOn(named("goDownloadDeps"))
 
-        args("install", "github.com/golang/mock/mockgen")
+        args("install", "google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest")
```
dezyh, line 86: Is it worth pinning a specific version rather than latest?

dezyh (approved):
I don't understand exactly what the more gradley bits are doing, but the protoc stuff looks correct. If it builds and we can consume it correctly, then I'm happy.

JamieCrisman:
The reason for manually doing the protoc: the curio plugin does not support multiple protoc plugins. protoc-gen-go split the grpc code gen into two different plugins. But again the gradle changes are get all the necessary binaries, remove mockgen, generate protos that's compatible with the current binaries. 

Go mod changes are bumping up the binding dependencies.

We'll need to do more work to bring curio up to date to update this in the plugin. But manually running it gets it working as well.

PR #331, opened 2025-05-02
fix: github actions/cache v1 no longer useable, update to current

JamieCrisman:
[v1 is no longer supported.](https://github.com/actions/toolkit/blob/1b1e81526b802d1d641911393281c2fb45ed5f11/packages/cache/README.md)


dsypniewski (approved):
LGTM

PR #332, opened 2025-05-02
golang: fix copy order to ensure gomod updates

JamieCrisman:
copy ordering was not being enforced. the generateProto will wipe the directory, so need to ensure the mod copies are after.


dsypniewski (approved):

dezyh (approved):
Amazing

